PR #14, opened 2020-09-24
Added GK for thermal

Fratorhe:
Hi, 

I added the GK computation for thermal conductivity.
This implementation uses the results from a compute heat/flux from LAMMPS to compute the thermal conductivity. This compute gives the flux j, so the autocorrelation can be computed and obtain k. 
For the moment, it is limited to Real LAMMPS units. I will add an issue with an idea to improve this.
I also did some re-formatting to the main classes (moving some methods from the Class Trajectory, to its parent Class Trajectory_Methods), such that my new class inherits also from Trajectory_Methods.


SamTov (changes requested):
The over arching design of the software is such that one can avoid printing specific properties for lammps and thus reducing the overall data requirements in large simulations. As far as I understand, the thermal conductivity calculation may be performed using only the forces, positions and velocities? If this is the case, I would rather have separate classes for a single property. What are your thoughts on this?

As a separate comment, the changes you have made to the main class structures are great and I would like to merge them, however we should discuss more how to include this thermal conductivity term.

SamTov (approved):
After looking through some added literature, it is preferable to also include potential energy corrections to the site, therefore I am okay with having a seperate class file for this specific calculation.

Fratorhe:
I am not sure what you mean with your last comment. We can discuss about tomorrow.

Issue #213, opened 2021-05-05
add_data does not detect file type and tries to always load lammps


PythonFZ:
**Describe the bug**

`test_project.add_data('./trajectory.extxyz')``
uses the Lammps file loader, even though it is clearly an `.extxyz` file

**Expected behavior**
We should try to distinguish the file type by the ending.

PythonFZ:
This also affects https://github.com/zincware/MDSuite/blob/8efed9524868b46a75f2c494d9ffbb37614d72ab/mdsuite/project.py#L218 because it is currently not possible to load `extxyz` data

SamTov:
It is a little more complicated in terms of a .xyz extension but certainly for .extxzy we can enforce this. There is no lammps trajectory extension and often people will write .xyz anyway.

In the case of the project class that is certainly problematic. I would propose a file identification module in the meta_functions module or perhaps better as a standalone in the file_io package.

PR #214, opened 2021-05-05
Add file format option to project class.

SamTov:
In response to #213 I have added a file format option to the project classes add_data method. No automatic file checking has been implemented as of yet.


mdsuite/project.py
```diff
@@ -225,7 +225,7 @@ def add_data(self, data_sets: dict):
         Updates the experiment classes.
         """
         for item in data_sets:
-            self.experiments[item].add_data(data_sets[item])
+            self.experiments[item].add_data(data_sets[item], file_format=file_format)
```
PythonFZ, line 229: This assumes that all files have the same file_format. Maybe also use a dictionary instead? We could use an `assert data_sets.keys() == file_format.keys()` to ensure that they are the same. Or we allow both, a single file format and a dictionary. Also docstrings.

PythonFZ (changes requested):

mdsuite/project.py
PythonFZ: Should be solved in fa11735

PythonFZ (approved):

Issue #210, opened 2021-05-04
RDF minibatching does seem to use less data than without.


PythonFZ:
**Describe the bug**
RDFs look way worse with minibatching enabled

**To Reproduce**
Run the RDF in `docs/source/_user_guide/molten_salt_example.ipynb`

PythonFZ:
fixed in 21c85cb537f8468fe214902f17d47cb21a359680

PR #216, opened 2021-05-05
fix Issue with RDF output

PythonFZ:
To solve #210 I added a return value to the RDF. Because RDF is set to be experimental, it now always returns the rdf dict which is bad for running it in Jupyter. I removed the return statement. I also changed the `get_machine_properties` because it raised a `ValueError` on Google Colab. I think, that changing the `raise Error` to a `log.warning` should solve that. It seems to work, but I don't know if it comes with other issues?


SamTov (approved):
The GPU issues seem to be persistent on any cluster-based system. We may need to look into a better means of GPU management.

PR #222, opened 2021-05-26
Change to new MSD computation approach.

SamTov:
I have update all MSD computations to use a new, better performing TF approach. Changes include:

* Introduce general MSD method in the Calculator class
* Update all MSD based calculators to use new approach. Note, not all of them use the method in the calculator class as many require small changes. We can consider a better method for consolidation but for now it is still only one line of code.
* Fixed the set charge method in the experiment class and included type hints in the parameters to better facilitate use of the method in interactive sessions.

I have tested the implementation on Molten NaCl and KCl and the results match that of the previous method but are significantly faster. I have done this for self-diffusion and ionic conductivity. This covers all new approaches but I have not explicitly tested thermal conductivity and viscosity.

### DIFF
--- a/mdsuite/calculators/einstein_helfand_ionic_conductivity.py
+++ b/mdsuite/calculators/einstein_helfand_ionic_conductivity.py
@@ -87,14 +87,18 @@ class EinsteinHelfandIonicConductivity(Calculator):
 
         Parameters
         ----------
-        experiment :  object
-            Experiment class to call from
         plot : bool
                 if true, plot the tensor_values
         data_range :
                 Number of configurations to use in each ensemble
         save :
                 If true, tensor_values will be saved after the analysis
+        correlation_time : int
+                Correlation time to use in the analysis.
+        export : bool
+                If true, export the results to a csv.
+        gpu : bool
+                If true, reduce memory usage to the maximum GPU capability.
         """
 
         # parse to the experiment class
@@ -105,7 +109,6 @@ class EinsteinHelfandIonicConductivity(Calculator):
         out = self.run_analysis()
 
         self.experiment.save_class()
-        # need to move save_class() to here, because it can't be done in the experiment any more!
 
         return out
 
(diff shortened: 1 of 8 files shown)


mdsuite/calculators/einstein_helfand_ionic_conductivity.py
```diff
@@ -161,8 +164,7 @@ def _apply_operation(self, ensemble, index):
         -------
         MSD of the tensor_values.
         """
-        msd = (ensemble - (
-            tf.repeat(tf.expand_dims(ensemble[0], 0), self.data_range, axis=0))) ** 2
+        msd = tf.math.squared_difference(ensemble, ensemble[None, 0])
```
PythonFZ, line 167: Why not use `Calculator._msd_operation`

mdsuite/calculators/einstein_helfand_thermal_conductivity.py
```diff
@@ -158,8 +158,8 @@ def _apply_operation(self, ensemble, index):
         -------
         MSD of the tensor_values.
         """
-        msd = (ensemble - (
-            tf.repeat(tf.expand_dims(ensemble[0], 0), self.data_range, axis=0))) ** 2
+        msd = tf.math.squared_difference(ensemble, ensemble[None, 0])
```
PythonFZ, line 161: Why not use `Calculator._msd_operation`

mdsuite/calculators/einstein_helfand_thermal_kinaci.py
```diff
@@ -159,8 +159,8 @@ def _apply_operation(self, ensemble, index):
         -------
         MSD of the tensor_values.
         """
-        msd = (ensemble - (
-            tf.repeat(tf.expand_dims(ensemble[0], 0), self.data_range, axis=0))) ** 2
```
PythonFZ, line 163: Why not use `Calculator._msd_operation`

PythonFZ (commented):
I can't say much about the experiment changes, but the rest looks good.
Is there a reason you aren't using `Calculator._msd_operation` everywhere?

mdsuite/calculators/einstein_helfand_ionic_conductivity.py
SamTov: Yes, the dimension expansion is along differing axis so the option is either to have the computation in the calculator itself, or to have quite a large logic segment in the main _msd_operation method. We would need logic for square=True/False and then a check if system=True logic section under each. It isn't bad but it didn't seem elegant. I can think of a way to integrate it all.

PythonFZ: Okay, in that case I don't see a reason, why it should be changed. Looks all good

PythonFZ (approved):

PR #224, opened 2021-05-26
Sql error fix

SamTov:
Fix issue #223 for Green-Kubo and Einstein distinct diffusion coefficients.

PR #225, opened 2021-05-26
Bugfix

PythonFZ:
- Fix an issue with the SQL Database for Distinct diffusion and GK #224 
- Improve the MSD computation #222


SamTov (approved):
Everything is in order. We can merge it as these are urgent fixes.

PR #228, opened 2021-05-27
Change dict parse to a deep copy.

SamTov:
Fixes an issue with consecutive loading of experiments. I didn't make an issue as I fixed it immediately.


PythonFZ (approved):
Looks good.

I never touched those files, so it might be a dumb question, but why are those global variables in the first place? Wouldn't it be more convenient to make them class attributes?

SamTov:
I think it would be better to store them in a data file and load them in as a dictionary

PythonFZ:
And in addition to that, there is also `var_names` in `extxyz_trajectory_reader.py` that may be affected?

PR #240, opened 2021-06-07
Remove Tutorials

PythonFZ:
Fix #239


SamTov (commented):
We should add some options for tutorials later on. For now can we also update the README to direct people to the docs for examples.

SamTov (approved):
Will approve now.

PR #243, opened 2021-06-08
Simulation data

### LINKED ISSUE #242: metadata for experiments
store metadata of any kind on experiments.
In the project, the values of the experiments should be accessible through the parameters of the experiments 

Then I don't need to do that manually

no idea

**User story**
I run 100 simulations, varying 2 obscure parameters like packing fraction and flagellar unbundling rate.
I the want a plot of the Diffusion coefficient vs unbundling rate, with a different line color for each packing fraction.
To do that I would like tabulated data like

exp_id | pack_f | unbundle_rate | ... [all other metadata like time step, Temperature, vox volume etc] ... | msd
0        | 0.2    |   16.4  | [...] | 2e-8
1        | 0.2    |   16.8  | [...]  | 2.5e-8
2        | 0.1    |   16.4  | [...]  | 7e6
.
.
.

PythonFZ:
Fix #242 

We introduce a method to save arbitrary user data.

- Introduce `mds.Experiment.simulation_data` dictionary, that can contain any json serializable data.
- Add a `simulation_data.yaml` file that contains all the user information


PythonFZ:
We can think about eventually removing the `save_file.write(pickle.dumps(self.__dict__))` and replace it with some class properties that read from a file on default.

SamTov (approved):
Looks good. We now need to update the Project class so that it can use this.

PR #251, opened 2021-06-22
Rdf comparison doc

PythonFZ:

* Add an example on how to compute RDFs of different systems and combine them in a single plot


PythonFZ:
I accidentally based this one on the ADF/RDF restructure - so I changed my PR

PR #252, opened 2021-06-22
fix norm factor

PythonFZ:

*  potential fix for the normalization factor 
image

The power `4` can be set to any value using `norm_power` which defaults to 4


SamTov (commented):
Looks nice but do you have an example image?

PythonFZ:
- Power = 8
image
- Power = 4
image
- Power = 2
image
- Power = 0
image
- Power = -4 
image


I guess you can get some distance / angle resolution out of that. All of them use a large cutoff = 8 A

PR #253, opened 2021-06-22
Arbitrary tau sam tov

### LINKED ISSUE #247: Allow for arbitrary tau selection in MSD calculations.
It would be helpful to have the ability to select not only a delta tau of 1 when performing an MSD calculation. That is, to compute the MSD at steps 0, 2, 4, ... rather than at each time-step.

It would become more flexible.

The user should be able to parse a taus argument which can be in integer (points spaced by n) or a list specificity exactly which points to take.

Code-wise the `msd_operation` method will need to be changed. It should suffice to change the 
```python
 if square:
      return tf.math.squared_difference(ensemble, ensemble[:, None, 0])
```
portion to 

```python
 if square:
      return tf.math.squared_difference(ensemble[:, indices], ensemble[:, None, 0])
```

SamTov:
Fix #247 

* Allo for arbitrary tau selection in the Einstein diffusion coefficients.

Run some einstein diffusion as follows.
```
experiment.run_computation.EinsteinDiffusionCoefficients(tau_values=np.s_[::3])  # take every third
experiment.run_computation.EinsteinDiffusionCoefficients(tau_values=11)  # take 11 points evenly spaced
experiment.run_computation.EinsteinDiffusionCoefficients(tau_values=[1, 12, 13, 18, 22, 46])  # take the steps in the list.
```

### DIFF
--- a/mdsuite/calculators/calculator.py
+++ b/mdsuite/calculators/calculator.py
@@ -378,7 +378,8 @@ class Calculator(metaclass=abc.ABCMeta):
                 Mean square displacement.
         """
         if square:
-            return tf.math.squared_difference(np.array(ensemble)[:, self.tau_values], ensemble[:, None, 0])
+            return tf.math.squared_difference(tf.gather(ensemble, self.tau_values, axis=1),
+                                              ensemble[:, None, 0])
         else:
             return tf.math.subtract(ensemble, ensemble[:, None, 0])
 
@@ -397,7 +398,8 @@ class Calculator(metaclass=abc.ABCMeta):
             self.data_resolution = len(self.tau_values)
             self.data_range = self.tau_values[-1] + 1
         if type(self.tau_values) is slice:
-            self.data_resolution = len(np.linspace(0, self.data_range, self.data_range)[self.tau_values])
+            self.tau_values = np.linspace(0, self.data_range - 1, self.data_range, dtype=int)[self.tau_values]
+            self.data_resolution = len(self.tau_values)
 
     def _prepare_managers(self, data_path: list):
         """
--- a/mdsuite/calculators/calculator.py
+++ b/mdsuite/calculators/calculator.py
@@ -33,7 +33,7 @@ from mdsuite.calculators.transformations_reference import switcher_transformatio
 from mdsuite.database.properties_database import PropertiesDatabase
 from mdsuite.database.analysis_database import AnalysisDatabase
 from tqdm import tqdm
-from typing import Union
+from typing import Union, List, Any
 
 log = logging.getLogger(__file__)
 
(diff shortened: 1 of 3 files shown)


mdsuite/calculators/calculator.py
```diff
@@ -364,10 +378,26 @@ def _msd_operation(ensemble: tf.Tensor, square: bool = True):
                 Mean square displacement.
         """
         if square:
-            return tf.math.squared_difference(ensemble, ensemble[:, None, 0])
+            return tf.math.squared_difference(np.array(ensemble)[:, self.tau_values], ensemble[:, None, 0])
```
PythonFZ, line 381: I'd suggest for better performance that we don't use numpy here.
If we need the indexing I'd suggest using `tf.gather` or `tf.gather_nd` for this

mdsuite/calculators/calculator.py
```diff
@@ -364,10 +378,26 @@ def _msd_operation(ensemble: tf.Tensor, square: bool = True):
...
+        -------
+        Updates the class state.
+        """
+        if type(self.tau_values) is int:
+            self.data_resolution = self.tau_values
+            self.tau_values = np.linspace(0, self.data_range - 1, int(self.tau_values), dtype=int)
+        if type(self.tau_values) is list:
+            self.data_resolution = len(self.tau_values)
+        if type(self.tau_values) is slice:
+            self.data_resolution = len(np.linspace(0, self.data_range, self.data_range)[self.tau_values])
```
PythonFZ, line 400: Isn't this just `len(np.arange(self.data_range)[self.tau_values])` 
I don't know the type slice, but I guess `len(self.tau_values)` does not work?

mdsuite/calculators/calculator.py
```diff
@@ -452,11 +482,14 @@ def _save_data(self, name: str, data: pd.DataFrame):
                 db[self.database_group].create_dataset(title, data=data, dtype=float)
         """
 
-    def _plot_fig(self, fig: matplotlib.figure.Figure, ax: Axes,
+    def _plot_fig(self,
+                  fig: matplotlib.figure.Figure,
+                  ax: Axes,
```
PythonFZ, line 488: Why import `Axes` but not `Figure` ?

PythonFZ (changes requested):

mdsuite/calculators/calculator.py
SamTov: no it's okay how it is. The data resolution is just a scaling factor to re-scale the time arrays or msd arrays. 

```python
time = np.linspace(0, self.data-range, self.data-resolution)
```

mdsuite/calculators/calculator.py
SamTov: Yeah this is okay but of the three outcomes only two require the gather, the np.s_ object works without numpy

mdsuite/calculators/calculator.py
SamTov: I didn't add this, i just reformatted it.

mdsuite/calculators/calculator.py
PythonFZ: Okay, I checked and

```py
 np.linspace(0, x, x - 1) == np.arange(x)
```

So I guess this is resolved then

mdsuite/calculators/calculator.py
SamTov: I will try to parse all the inputs as np.s_ objects, it could work.

SamTov: Or can try to parse the np.s_ object to the gather. The other approach failed.

SamTov: Okay so I can just make the slices arrays and then gather works fine.

SamTov (commented):
Updated all requests to completely use TensorFlow.  Code has been tested for all slice types and on a large system of emimF

PythonFZ (approved):

PR #256, opened 2021-06-24
catch empty sim data file

christophlohrmann:
an empty sim data file can occur if the dict passed in the setter cannot be written (e.g. it contains an ``np.array``).


PythonFZ:
That should not happen. Thanks for the PR. I'll merge it.

PR #258, opened 2021-06-24
remove all matplotlib style changes

### LINKED ISSUE #246: mdsuite changes matplotlib style
**Describe the bug**
When importing mdsuite, the matplotlib style is set to 'classic'.

**To Reproduce**
plot anything with matplotlib after ``import mdsuite`` and you will get the 'classic' style.

**Expected behavior**
mdsuite not to change global states of other modules.
when plotting, ``mdsuite`` should either use the user settings or revert to them when it is done.

``matplotlib.style.use('default')`` after ``import mdsuite`` brings me back to what I want, but I'd rather not do that everytime

christophlohrmann:
Fix #246 

* Remove all changes of the matplotlib style global

* this PR closes #246 as an issue, but not the discussed changes to mdsuite plotting philosophy


SamTov (approved):
No more classics. Looks good, now we have to push harder on the development of our plotting class.

PR #261, opened 2021-06-29
Optimize the RDF code to better work with larger systems

### LINKED ISSUE #260: Improving RDF
**Describe the bug**
We need to improve the species sorting of the RDF calculators.

**Some ideas**

All of the discussion here is on the `compute_species_values` method beginning line 569 if the `radial_distribution_function.py` module in the development branch

* Line 589 `particles_list = [len(self.experiment.species[item]['indices']) for item in self.experiment.species]` should be changed to only the calculator species. This would result in less loops for specific species systems.
* Line 394 `cutoff_mask = tf.cast(tf.less(tensor, cutoff), dtype=tf.bool)  # Construct the mask` The bool casting here is irrelevant. 
*  `mini_get_pair_indices` and `particle_list` should only need to be computed once but are computed in every iteration.

PythonFZ:
Fix #260 

* Using `tf.sparse` and more `tf.function`

Run any RDF code - preferably with `logging.DEBUG` to see the excat times.

This method perfoms way better for small systems and should be slightly better for large systems.
Unfortunately the `tf.sparse.slice` is not very fast and doesn't seem to be fully parallized which leads to some performance bottlenecks.

### DIFF
--- a/mdsuite/calculators/radial_distribution_function.py
+++ b/mdsuite/calculators/radial_distribution_function.py
@@ -521,22 +521,34 @@ class RadialDistributionFunction(Calculator, ABC):
         tf.Tensor, str
 
         """
+
+        startx = timer()
         indices = tf.transpose(indices)
-        background = tf.tensor_scatter_nd_update(tf.fill((atoms_per_batch, n_atoms), -1),
-                                                 indices,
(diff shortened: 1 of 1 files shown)


PythonFZ:
+ check what was wrong with #266

PR #264, opened 2021-07-04
check previous files

### LINKED ISSUE #257: Add support for not adding the same file path twice.
If you accidentally forget to comment out a trajectory file addition you will duplicate data in the database. We should add an exact file path check to avoid this.

It would prevent duplicated data from an identical file path

We will need to have some file to store the data such as a hidden .file_paths.txt file. This file should be stored in the storage path along with the experiment data. The experiment should then check if the file path given matches one of the paths in this hidden file exactly. There should also be a force_add argument for when one wishes to add the data anyway. See below for details.

```python
# Line 368 experiment.py on the development branch
def add_data(..., force_add: bool = False):
    """
    ....
    ....
    force_add : bool (default = False)
            If true, data is forcefully added whether it has been seen or not.
    """
    ... # Check if the trajectory file is None -- already exists
    # Start before the load trajectory reader line -- name it anything
    if self.check_previous_addition():
        log.log("Already added")
        return
```

Somewhere in `experiment.py`

```python
# New method required.
def check_previous_file():
    """
    Check to see if the file being added has already been seen
    """
    data = open('hidden_file.txt').read()  # read in the file
    if self.trajectory_file in data:
        return True
    else:
        return False
```

SamTov:
Fix #257 

* Add method to read/write a hidden file that stores information about all previously read in files.
* Will end the add_data method if a file has already been read.
* Added keyword argument `force` to allow for a file to be forcefully added.

```python
experiment.add_data('test.xyz')

experiment.add_data('test.xyz')

experiment.add_data('test.xyz', force=True)
```

### DIFF
--- a/mdsuite/experiment/experiment.py
+++ b/mdsuite/experiment/experiment.py
@@ -364,8 +364,37 @@ class Experiment:
 
         return attributes
 
-    def add_data(self, trajectory_file: str = None, file_format: str = 'lammps_traj', rename_cols: dict = None,
-                 sort: bool = False):
+    def _check_read_files(self, file_path):
+        """
+        Check if a file has been read before and add it to the hidden file.
+
+        Parameters
+        ----------
+        file_path : str
+                Path to the file.
+
+        Returns
+        -------
+
+        """
+        with open('.read_files.txt', 'a') as f:
+            data = []
+            for line in f:
+                data.append(f.readline())
+
+            result = file_path in data  # check if it exists.
+            if result:
+                return result
+            else:
+                f.write(f'{file_path}\n')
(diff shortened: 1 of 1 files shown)


PythonFZ (approved):

PythonFZ:
Looks good, altough we could also consider putting that into the database to have everything in one place?

SamTov:
Yeah I thought this information to be too insignificant for the database.I guess it is probably a good place for it though. Before doing this we should develop some nice database schema so that we do not need to re-factor this in the future. For now I will merge it as is. When we change the database we can update the check method to handle this.

Issue #233, opened 2021-05-31
Plot Energies/Forces/... over time


PythonFZ:
**What feature would you like to see added?**
For the investigation of ML based trajectories it can be interesting to be able to look at properties such as energy, forces, densities or velocities over time. This might become handy if a ML trajectory explodes. Because MDSuite has access to all this properties it should be possible to visualize them.

**Describe how this feature would work/possible implementations**
I'd propose an implementation similar to the calculators. Something like

```py
experiment.visualize(property="energy", smoothing=0.2, start=0, stop=-1, step=1)
# property could also be list, in case for force_x, force_y, force_z ? 

experiment.visualize.energy(smoothing=0.)
```

I would also like to be able to export that data easily into numpy arrays, so we could either add that directly as function or use the existing properties database for that (but it might not be sensible to store the data in SQL and  H5Py simultaneously)

**Additional context**
- Because MDSuite should know all the units it should be possible to label the axis correctly.
- How should we handle properties like forces, have `force_x, force_y, force_z` and `force` gives the magnitude?

SamTov:
This is a great idea, not only for the ML potentials, but also for for any benchmark of a potential as you always want to ensure you get the correct density.

I would be interested in having a calculator to do this sort of thing, or adding it to a visualisation module. My argument for a calculator is that we can compute things like average temp or density. It would also reduce confusion with the visualiser that we are implementing for the atoms.

**Additional Notes**
In the cases of vectored properties, it is probably rare that we want to visualise them in a time series at the same time. We could consider making volumes or heat maps, which would definitely come under visualisers as this is the direction we want to take with things like spatial distribution functions, but this requires additional discussion.

**Implementation**
I would suggest we start by having a scalar value calculator which studies scalars such as density, volume, force magnitude, and so on. This could go in the calculator package and have the exact parameters you suggested with perhaps some additional methods for computing values.

In the case of vector fields and multi-dimensional plotting, we should consider adding a visualisation module for these things and putting some effort into studying how we can best represent them.

SamTov:
How did this go?

PythonFZ:
`exp.analyse_time_series.Energies()` should work fine. 
I haven't implemented anything exceeding the energies.

SamTov:
Maybe a better thing here is to have some profiling module. We could have a binning or plotting argument that could handle lots of stuff. I would also include particle density profiles here too. @PythonFZ

PythonFZ:
I like the idea, how could this look like?

[...]

PR #267, opened 2021-07-06
Time series

PythonFZ:
Start looking into #233 

*  Added `exp.analyse_time_series.Energies()`

1. Create some experiment
2.  Run
```py
exp.analyse_time_series.Energies()
```

### Additional changes
- default the load data to the full trajectory
- optimize the `RunCompution` to make it more general
- add an exclude statement on the pickle class dict loading

### DIFF
--- a/mdsuite/database/simulation_database.py
+++ b/mdsuite/database/simulation_database.py
@@ -424,7 +424,7 @@ class Database:
                 if item in mapping:
                     db.move(item, mapping[item])
 
-    def load_data(self, path_list: list = None, select_slice: np.s_ = None, dictionary: bool = False,
+    def load_data(self, path_list: list = None, select_slice: np.s_ = np.s_[:], dictionary: bool = False,
                   scaling: list = None, d_size: int = None):
         """
         Load tensor_values from the database_path for some operation.
--- a/mdsuite/experiment/experiment.py
+++ b/mdsuite/experiment/experiment.py
@@ -27,6 +27,7 @@ from tqdm import tqdm
 import importlib.resources
 from datetime import datetime
 from mdsuite.calculators.computations_dict import dict_classes_computations, dict_classes_db
+from mdsuite.time_series import time_series_dict
 from mdsuite.transformations.transformation_dict import transformations_dict
 from mdsuite.file_io.file_io_dict import dict_file_io
 from mdsuite.utils.units import units_dict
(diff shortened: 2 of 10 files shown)


mdsuite/database/simulation_database.py
```diff
@@ -424,7 +424,7 @@ def change_key_names(self, mapping: dict):
                 if item in mapping:
                     db.move(item, mapping[item])
 
-    def load_data(self, path_list: list = None, select_slice: np.s_ = None, dictionary: bool = False,
+    def load_data(self, path_list: list = None, select_slice: np.s_ = np.s_[:], dictionary: bool = False,
```
SamTov, line 427: We need to keep this in mind during our testing phase as there could be some unwanted results.

mdsuite/experiment/experiment.py
```diff
@@ -40,6 +41,49 @@
 log = logging.getLogger(__file__)
 
 
+class RunModule:
```
SamTov, line 44: This is nice but can we move it to a separate module and import it? leave it within the experiment directory but make a new file. We can of course discuss this but my thoughts are that we separate modules by their directory structure and so we can spread out class declaration within these directories. As the experiment class is very big and has a lot to do, it might be best to move the computation handling to its own module.

mdsuite/time_series/base.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @property
+    def database(self):
+        """Get the database"""
+        if self._database is None:
+            self._database = Database(name=self.experiment.database_path + r"\database.hdf5")
+        return self._database
+
+    @property
+    def data(self):
```
SamTov, line 54: I think a species argument would be better. Add species as a parameters and set all species as default.

mdsuite/time_series/base.py
```diff
@@ -0,0 +1,70 @@
...
+                [self.database.load_data([f'{species}/{self.loaded_property}']) for species in self.experiment.species],
+                axis=0
+            )
+        return self._data
+
+    def plot(self):
+        """Plot the data over timesteps"""
+        fig, ax = plt.subplots()
+        ax.plot(tf.einsum("atx -> t", self.data))
+        # performe a reduce sum over atoms "a" and property dimension "x" to yield time steps "t"
```
SamTov, line 67: perform*

SamTov (changes requested):
Should we also offer some options for taking running averages?

mdsuite/experiment/experiment.py
PythonFZ: Yeah, sure. I also thought about it, where do you think it would fit? Probably into utils?

SamTov: I would still leave it in the experiment directory. Just move it to its own module.

PythonFZ: fixed in d17c8a1c53dd9ed64ebcaafd63a498cc3198f412

SamTov (changes requested):
Have you added the additional modules to the documentation pages?

SamTov (approved):
Looks perfect.

PR #268, opened 2021-07-07
Speed up the RDF code

PythonFZ:

* make the RDF code faster
* applying `tf.function` is possible
* add debug message on atom pairs per second

run an RDF

### DIFF
--- a/mdsuite/calculators/calculator.py
+++ b/mdsuite/calculators/calculator.py
@@ -38,6 +38,7 @@ from tqdm import tqdm
 from typing import Union, List, Any
 
 from typing import TYPE_CHECKING
+
 if TYPE_CHECKING:
     from mdsuite import Experiment
 
@@ -145,6 +146,9 @@ class Calculator(metaclass=abc.ABCMeta):
         self.database_group = None
         self.analysis_name = None
 
+        # Properties
+        self._dtype = tf.float64
+
         # Prevent $DISPLAY warnings on clusters.
         if self.experiment.cluster_mode:
             import matplotlib
@@ -156,7 +160,7 @@ class Calculator(metaclass=abc.ABCMeta):
                          data_range: int = 500,
                          correlation_time: int = 1,
                          atom_selection: object = np.s_[:],
-                         tau_values : Union[int, List, Any] = np.s_[:],
+                         tau_values: Union[int, List, Any] = np.s_[:],
                          export: bool = True,
                          gpu: bool = False):
         """
@@ -794,3 +798,8 @@ class Calculator(metaclass=abc.ABCMeta):
             pass
         else:
             return self.perform_computation()
+
+    @property
+    def dtype(self):
+        """Get the dtype used for the calculator"""
+        return self._dtype
--- a/mdsuite/calculators/calculator.py
+++ b/mdsuite/calculators/calculator.py
@@ -12,6 +12,8 @@ Parent class for different analysis
 Summary
 -------
 """
+from __future__ import annotations
+
 import logging
 import numpy as np
 import os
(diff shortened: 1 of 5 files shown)


mdsuite/calculators/radial_distribution_function.py
```diff
@@ -575,14 +540,27 @@ def compute_species_values(indices, atoms_per_batch, start, d_ij):
...
+                start_ = (sum(particles_list[:tuples[0]]) - start_batch, sum(particles_list[:tuples[1]]))
+                size_ = (
+                    particles_list[tuples[0]],
+                    particles_list[tuples[1]]
+                )
+
+                names = self._get_species_names(tuples)
+
+                # select the indices that are within the boundaries of the current species / molecule
+                mask_1 = tf.math.logical_and(indices[:, 0] > start_[0], indices[:, 0] < start_[0] + size_[0])
```
SamTov, line 555: This is fast, but is there a faster way without the logic? I am not sure how tf handles logic but it is typically slower than building a binary mask.

SamTov (commented):
Looks great.

mdsuite/calculators/radial_distribution_function.py
PythonFZ: I don't know.
```py
mask_1 = (indices[:, 0] > start[0]) & (indices[:, 0] < stop[0])
mask_2 = (indices[:, 1] > start[1]) & (indices[:, 1] < stop[1])
```

This is just another notation that yields the same speeds I tried

mdsuite/calculators/calculator.py
```diff
@@ -35,6 +37,10 @@
 from tqdm import tqdm
 from typing import Union, List, Any
 
+from typing import TYPE_CHECKING
+if TYPE_CHECKING:
```
SamTov, line 41: This always causes me issues. I am not sure we should do it.

mdsuite/calculators/radial_distribution_function.py
```diff
@@ -33,9 +34,14 @@
 
 from mdsuite.calculators.calculator import Calculator
 from mdsuite.utils.meta_functions import split_array
+from mdsuite.utils.linalg import apply_minimum_image, get_partial_triu_indices
 
 from timeit import default_timer as timer
 
+from typing import TYPE_CHECKING
+if TYPE_CHECKING:
+    from mdsuite import Experiment
```
SamTov, line 43: Same comment

SamTov (approved):
My only question is about the Experiment type checking stuff. It seems to cause issue for me sometimes but if it is running then it's okay.

mdsuite/calculators/calculator.py
PythonFZ: I understand what you mean, but it helps the type checking and with `from __future__ import annotations` this should be perfectly fine. The downside of having this import looking suspiciouse I think is less important than having the benefti of correct type checking for circular imports.

PythonFZ:
> My only question is about the Experiment type checking stuff. It seems to cause issue for me sometimes but if it is running then it's okay.

Could you maybe try to run it on your setup?
Maybe with a 3 or more component system if you have one ready, to check that it really works.
If that is resolved I'll merge it.

PythonFZ:
One thing we need to make sure is that the dtypes are set correctly. This is currently missing. I'll suggest adding a `self.dtype` attribute to the calculator base class

PythonFZ:
Maybe perform a double batch on the RDF data in the minibatch loop, that way we could increase the batch size, because the data gets smaller in the loop and it could speed it up. This might be more difficult with the memory management and TQDM.

SamTov (commented):
It looks good but have you tested it on > 2 species?

PythonFZ:
> It looks good but have you tested it on > 2 species?

I have not. I don't have a system on hand to test it

SamTov:
tested IL. All works we can push

PR #270, opened 2021-07-09
Visualizer addition.

SamTov:

* Added a visualizer module
* Currently it cannot do much but I would like to have the updated code in the dev branch.

```
import mdsuite as mds

experiment  = mds.Experiment(...)
experiment.add_data(...)

experiment.run_visualization()
```

#### Notes

The visualizer will just display the first configuration of the data. If you try to run the dynamic visualization it will fail. I will add functionality for selecting which configuration to view but for now you can render quite a nice image, dynamically change background colours and lighting properties and then save the image.


SamTov:
I guess we should leave this open until a few more methods are added but feel free to read it.

PythonFZ:
> I guess we should leave this open until a few more methods are added but feel free to read it.

Maybe convert it to a draft PR then?

mdsuite/visualizer/visualizer.py
```diff
@@ -0,0 +1,229 @@
...
+        """
+        Add colours and sizes to species.
+
+        Returns
+        -------
+        Updates the class attributes.
+        """
+        data_name = "mdsuite.data"
+        json_name = 'PubChemElements_all.json'
+        with importlib.resources.open_text(data_name, json_name) as json_file:
```
PythonFZ, line 78: I just curious, are there any benefits over a plain `with open()` or `Path(data_name, json_name).read_text()` ?

SamTov: i didn't write any of this stuff so have no comment. I think it was Marco.

PythonFZ: Probably just another way to write it. It works, so resolved.

PR #274, opened 2021-07-14
add load_data

### LINKED ISSUE #272: Make reading from the SQL database easier
It could be a nice idea to have the sql database query built inside of the calculator that generated the data. This means that all of the data used in the query is already there and users do not need to learn a new set of interface rules to make that query.

It would make the code a lot more usable when it comes to exporting data

**Possible implementation**
We should have a `load=True` or something in each calculator that if it is set, the calculator tries to built a query and return the data rather than run the computation. If the data isn't there it could just run the analysis first, add it to the database and then return it.

PythonFZ:
Fix #272 

*  add `experiment.load_data`

```py
experiment.load_data.RadialDistributionFunction()
```

### DIFF
--- a/mdsuite/calculators/calculator.py
+++ b/mdsuite/calculators/calculator.py
@@ -80,7 +80,8 @@ class Calculator(metaclass=abc.ABCMeta):
                  correlation_time: int = 1,
                  atom_selection: object = np.s_[:],
                  export: bool = True,
-                 gpu: bool = False):
+                 gpu: bool = False,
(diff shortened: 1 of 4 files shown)


PythonFZ:
This will probably be replaced by changes made in #280

SamTov (approved):
This is okay and it will be over-written by new changes. However, if we introduce some functionality for one calculator it would be good practice to do it for them all. Or we should move it correctly into the parent class.

PR #279, opened 2021-07-19
adapt gs function and test

### LINKED ISSUE #227: Golden-Section search appears to converge slightly off.
**Describe the bug**
The golden section search appears to converge slightly off the correct values. It seems to under-predict the values.

**To Reproduce**
```python
import numpy as np
from mdsuite.utils.meta_functions import golden_section_search
def func(x: np.ndarray):
    return x**2

x = np.linspace(-10, 10, 100)

interval = golden_section_search([x, data(x)], -5, 5)
print(interval) # (-0.101, -0.303)
```

**Expected behavior**
The interval should be approximately (0.101,-0.101)

SamTov:
Fix #227 but not completely.

* Made the GS search recursive
* Updated the test so that it now runs and passes

This issue is not fixed. The algorithm is implemented correctly and does begin to converge for excellent resolution but due to use of discrete data it seems that some error is introduced and propagated such that there is always a shift in the the true minimum. I would say now we need to find a new approach altogether.

```bash
cd MDSuite/CI
pytest
```

### DIFF
--- a/CI/unit_tests/utils/test_meta_functions.py
+++ b/CI/unit_tests/utils/test_meta_functions.py
@@ -7,9 +7,11 @@ Copyright Contributors to the MDSuite Project.
 """
 import unittest
 from mdsuite.utils.meta_functions import join_path, get_dimensionality, get_machine_properties, \
-    line_counter, optimize_batch_size, linear_fitting_function, simple_file_read, round_down, split_array, find_item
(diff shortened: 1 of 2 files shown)


SamTov (commented):
Small changes and introduction of a recursive equation.

PythonFZ (approved):

PR #281, opened 2021-07-20
Fix the golden sector test

### LINKED ISSUE #227: Golden-Section search appears to converge slightly off.

SamTov:
Closes #227

* Adjusted tolerance on gs test. 
* Added tolerance as argument to the gs approach

```
cd MDSuite/CI
pytest
```

### DIFF
--- a/CI/unit_tests/utils/test_meta_functions.py
+++ b/CI/unit_tests/utils/test_meta_functions.py
@@ -141,9 +141,10 @@ class TestMetaFunction(unittest.TestCase):
             return x**2
         x_dat = np.linspace(-10, 10, 1000)
         data = [x_dat, func(x_dat)]
-        output = golden_section_search(data, 10, -10)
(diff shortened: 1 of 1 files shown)


PythonFZ (approved):

SamTov (commented):
No we just have to either add some arguments for this parameter in the calculators or automatically compute the grid resolution.

PR #285, opened 2021-07-23
Apply black coding style

PythonFZ:

* apply `black .`

We can keep this open and decide whether we want to use `black` coding style. 
Here we can see, how it would look like

See https://black.readthedocs.io/en/stable/ for more information


SamTov (approved):
All looks good. I like the idea of having black in the code.

PythonFZ:
moved over to #286

PR #286, opened 2021-07-23
add black check

PythonFZ:
Add a Github Action that checks, if the code fulfills the `black` requirements.


SamTov (approved):
We should also have a black config file in the repo to define the parameters of the checks. This also has code-coverage support so we should mention that.

Issue #229, opened 2021-05-27
Move all global variables to data file.


SamTov:
**What feature would you like to see added?**
In line with the discussion in PR #228, we should move all global variables in the file loading to a data file. 

**Describe how this feature would improve the code**
It will prevent any issues with global variables. This is particularly important when change are made to them during some analysis, as takes place in the file loading.

**Describe how this feature would work/possible implementations**
Move the dictionary variables to a JSON file and load them when they are required, i.e., in a class __init__.

PythonFZ:
As a possible alternative or addition we can use dataclasses.
An example for the units could be:

from dataclasses import dataclass

```py
from dataclasses import dataclass

@dataclass(frozen=True)
class Units:
    pressure: float = 1.23456
    length: float = 1.1234656
    
   def gk_prefactor(self):
       return self.pressure ** 2 * self.length ** 3 + ...
```

They can be read from a json file or written to a json file easily.

PythonFZ:
I think this is resolved? @SamTov

PR #296, opened 2021-08-12
Update GK usage

### LINKED ISSUE #295: Fix GK integration and move to TF
We should move away from using zero-padded acfs and just use the TF implementation. This would require that we also have an integration range argument so that one can compute the acf over a longer range than they integrate. This is not only more correct but it means we can immediately use TF for the implementation and get that GPU performance we are looking for.

It would be more scientifically correct and more flexible.

Change the acf implementation and include an integration_range argument.

SamTov:
Fix #295 

Moved the gk implementation to tensor flow so that it now runs in parallel and updated the interface.

* Updated the interface to allow for a custom integration range argument.
* Implemented auto-correlation in tensor flow
* Updated plotting to show the integration line with respect to the acf

#### Not addressed

* Correct implementation of the distinct coefficients. I still need to better understand how the simple TF correlation compares with scipy
* Correct test runs. Due to the introduction of pure acf computations we no longer have zero-padding. While this makes the code more accurate and flexible, we now will always required a lot of statistics to get converged results, even on fast systems like the molten salts. Therefore, we will need to update the data used in the tests to address this.

### DIFF
--- a/mdsuite/calculators/green_kubo_distinct_diffusion_coefficients.py
+++ b/mdsuite/calculators/green_kubo_distinct_diffusion_coefficients.py
@@ -65,20 +65,27 @@ class GreenKuboDistinctDiffusionCoefficients(Calculator):
         """
         super().__init__(experiment)
 
-        self.scale_function = {'linear': {'scale_factor': 5}}
-        self.loaded_property = 'Velocities'
+        self.scale_function = {"linear": {"scale_factor": 5}}
+        self.loaded_property = "Velocities"
 
-        self.database_group = 'Diffusion_Coefficients'
-        self.x_label = 'Time $(s)$'
-        self.y_label = 'VACF $(m^{2}/s^{2})$'
-        self.analysis_name = 'Green_Kubo_Distinct_Diffusion_Coefficients'
+        self.database_group = "Diffusion_Coefficients"
+        self.x_label = "Time $(s)$"
+        self.y_label = "VACF $(m^{2}/s^{2})$"
+        self.analysis_name = "Green_Kubo_Distinct_Diffusion_Coefficients"
         self.experimental = True
 
-    def __call__(self, plot: bool = False, species: list = None,
-                 data_range: int = 500, save: bool = True,
-                 correlation_time: int = 1, export: bool = False,
-                 atom_selection: dict = np.s_[:], gpu: bool = False,
-                 integration_range: int = None):
+    def __call__(
+        self,
+        plot: bool = False,
+        species: list = None,
+        data_range: int = 500,
+        save: bool = True,
+        correlation_time: int = 1,
+        export: bool = False,
+        atom_selection: dict = np.s_[:],
+        gpu: bool = False,
+        integration_range: int = None,
+    ):
         """
         Constructor for the Green Kubo diffusion coefficients class.
 
@@ -93,10 +100,15 @@ class GreenKuboDistinctDiffusionCoefficients(Calculator):
         save :
                 If true, tensor_values will be saved after the analysis
         """
-        self.update_user_args(plot=plot, data_range=data_range, save=save,
-                              correlation_time=correlation_time,
-                              atom_selection=atom_selection, export=export,
-                              gpu=gpu)
+        self.update_user_args(
+            plot=plot,
+            data_range=data_range,
+            save=save,
+            correlation_time=correlation_time,
+            atom_selection=atom_selection,
+            export=export,
+            gpu=gpu,
+        )
 
         self.species = species  # Which species to calculate for
 
@@ -114,7 +126,8 @@ class GreenKuboDistinctDiffusionCoefficients(Calculator):
             self.species = list(self.experiment.species)
 
         self.combinations = list(
-            itertools.combinations_with_replacement(self.species, 2))
+            itertools.combinations_with_replacement(self.species, 2)
+        )
 
         out = self.run_analysis()
 
@@ -136,8 +149,9 @@ class GreenKuboDistinctDiffusionCoefficients(Calculator):
         -------
         updates the class state
         """
-        for ensemble in tqdm(range(self.ensemble_loop), ncols=70,
-                             desc=str(combination)):
+        for ensemble in tqdm(
+            range(self.ensemble_loop), ncols=70, desc=str(combination)
+        ):
             self.vacf = np.zeros(self.data_range)
             start = ensemble * self.correlation_time
             stop = start + self.data_range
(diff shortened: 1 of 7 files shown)


mdsuite/calculators/green_kubo_distinct_diffusion_coefficients.py
```diff
@@ -152,15 +164,18 @@ def run_experimental_analysis(self):
...
             self.atom_selection = select_atoms
         for combination in self.combinations:
             type_spec = {}
             data_path = [join_path(item, 'Velocities') for item in combination]
             self._prepare_managers(data_path=data_path)
             type_spec = self._update_species_type_dict(type_spec, data_path, 3)
-            type_spec[str.encode('data_size')] = tf.TensorSpec(None, dtype=tf.int16)
-            batch_generator, batch_generator_args = self.data_manager.batch_generator(dictionary=True)
+            type_spec[str.encode('data_size')] = tf.TensorSpec(None,
+                                                               dtype=tf.int16)
```
PythonFZ, line 176: we should start using `self.dtype` - but here we are really going for int16? I don't thing int32 would be slower but safer longterm?

mdsuite/calculators/green_kubo_distinct_diffusion_coefficients.py
```diff
@@ -231,27 +250,36 @@ def _post_operation_processes(self, species: Union[str, tuple] = None):
...
         self._update_properties_file(properties)
         # Update the plot if required
         if self.plot:
-            plt.plot(np.array(self.time) * self.experiment.units['time'], self.vacf, label=species)
+            plt.plot(np.array(self.time) * self.experiment.units['time'],
+                     self.vacf,
+                     label=species)
+            plt.vlines((np.array(self.time) * self.experiment.units['time'])[
+                           self.integration_range], min(self.vacf),
+                       max(self.vacf))
```
PythonFZ, line 265: I guess this will be changed to the visualization module eventually?

mdsuite/calculators/green_kubo_ionic_conductivity.py
```diff
@@ -113,8 +123,10 @@ def _update_output_signatures(self):
         -------
 
         """
-        self.batch_output_signature = tf.TensorSpec(shape=(self.batch_size, 3), dtype=tf.float64)
-        self.ensemble_output_signature = tf.TensorSpec(shape=(self.data_range, 3), dtype=tf.float64)
+        self.batch_output_signature = tf.TensorSpec(shape=(self.batch_size, 3),
+                                                    dtype=tf.float64)
```
PythonFZ, line 127: `self.dtype` should be a property that can be set to `dtype=tf.float32` or `float64` so we can change the accuracy of the calculators for performance reasons

mdsuite/calculators/green_kubo_ionic_conductivity.py
```diff
@@ -155,11 +167,14 @@ def _apply_operation(self, ensemble, index):
...
-                                     method='auto') for idx in range(3)])
-        self.jacf += jacf[int(self.data_range - 1):]
-        self.sigma.append(np.trapz(jacf[int(self.data_range - 1):], x=self.time))
+        jacf = self.data_range * tf.reduce_sum(
+            tfp.stats.auto_correlation(ensemble,
+                                       normalize=False,
+                                       axis=0,
+                                       center=False), axis=-1)
+        self.jacf += jacf
+        self.sigma.append(np.trapz(jacf[:self.integration_range],
```
PythonFZ, line 176: `tfp.math.trapz` could be used, so we can eventually move everything into graph mode with tf.function

mdsuite/calculators/green_kubo_viscosity.py
```diff
@@ -119,13 +132,17 @@ def _calculate_prefactor(self, species: str = None):
...
-            self.experiment.units['time'] / self.experiment.units['energy']
+        denominator = 3 * (self.data_range - 1) * \
+                      self.experiment.temperature * \
+                      self.experiment.units['boltzman'] * \
+                      self.experiment.volume
+
+        prefactor_units = self.experiment.units['pressure'] ** 2 * \
+                          self.experiment.units['length'] ** 3 * \
+                          self.experiment.units['time'] / \
+                          self.experiment.units['energy']
```
PythonFZ, line 145: we should make this a dataclass or a named tuple:

```py
from dataclasses import dataclass

@dataclass(frozen=True)
class Units:
    pressure: float = 1.23456
    length: float = 1.1234656
    
   def gk_prefactor(self):
       return self.pressure ** 2 * self.length ** 3 + ...
```

so we can have `self.experiment.units.pressure` or `self.experiment.units.gk_prefactor()`

PythonFZ (approved):
I have not tested it, but therefore we do have the tests running.

Looks all good, I have a small comment on especially the `self.experiment.units` dict, that we could consider changing, but that is not really related to this PR. If we agree on it, we should make it an issue.

mdsuite/calculators/green_kubo_distinct_diffusion_coefficients.py
SamTov: absolutely. I will update that.

mdsuite/calculators/green_kubo_distinct_diffusion_coefficients.py
SamTov: Yes. In fact the idea is to have a trigger in the plotting that allows for a re-calculation of the integral by user input.

mdsuite/calculators/green_kubo_ionic_conductivity.py
SamTov: Sure. When was this introduced? Because I saw the code when I was updating things but had no idea it was introduced. This is the kind of thing we need to have in the dev docs or something as without information about it it will be hard to always enforce.

mdsuite/calculators/green_kubo_viscosity.py
SamTov: Issue #229 might be a good place for this kind of discussion. I suggest there that we push globals to a data file but a class also sounds quite good. We should discuss the pros and cons. In general though this is a really good idea.

mdsuite/calculators/green_kubo_ionic_conductivity.py
SamTov: Let's open an issue to address this.

mdsuite/calculators/green_kubo_ionic_conductivity.py
SamTov: This is only possible in the newest version of tfp which is not yet full deployed for all architectures.

Issue #312, opened 2021-09-15
Make unit tests easy


christophlohrmann:
by introducing mock experiments/projects where one can insert test data that is needed for the calculator/transformation that is tested

edit: another option is to write a file reader that accepts numpy arrays and the needed meta information so we can populate a small experiment for testing purposes without having to write the test data to a file

PR #314, opened 2021-09-16
Draft: Vel from pos

christophlohrmann:
Draft until #312 is done and I can test it properly


christophlohrmann:
this transformation does not use batching/generators (same as with wrap_coordinates). But there is so much code duplication because all transformators call their own generators and do their own species list and existance-checking that I do not want to add to that until the transformators get cleaned up

PR #326, opened 2021-09-17
Fix issues

### LINKED ISSUE #322: load data returns
1. project.load_data: always return dict (even if there is only one exp)
2. exp.load_data: list of computations. if one calculator does 2 calculations (e.g. diff coeff + msd), list(Computation), where computation holds the individual results (e.g. computation['coeff'], computation['msd'])

PythonFZ:
Fix #322 

* introduce `Parameter` for a structured input file to the database

### Open ToDos

Update all calculators to use the `Parameters` (Highlighted through a depreciation warning)

```

```

### DIFF
--- a/mdsuite/database/calculator_database.py
+++ b/mdsuite/database/calculator_database.py
@@ -127,9 +127,10 @@ class CalculatorDatabase:
     # TODO rename and potentially move to a RDF based parent class
     def _get_rdf_data(self) -> List[db.Computation]:
         """Fill the data_files list with filenames of the rdf tensor_values"""
+        # TODO replace with exp.load.RDF()
         with self.experiment.project.session as ses:
             computations = ses.query(db.Computation).filter(
-                db.Computation.computation_attributes.any(str_value="RDF", name="Property")
+                db.Computation.computation_attributes.any(str_value="Radial_Distribution_Function", name="Property")
             ).all()
 
             for computation in computations:
--- a/CI/integration_tests/calculators/_test_einstein_distinct_diffusion_coefficients.py
+++ b/CI/integration_tests/calculators/_test_einstein_distinct_diffusion_coefficients.py
@@ -69,9 +69,9 @@ def test_eddc_project(traj_files, true_values, tmp_path):
     project = mds.Project()
     project.add_experiment("NaCl", data=traj_files[0], timestep=0.002, temperature=1400)
 
-    project.run_computation.EinsteinDistinctDiffusionCoefficients(plot=False, data_range=300, correlation_time=1)
+    project.run.EinsteinDistinctDiffusionCoefficients(plot=False, data_range=300, correlation_time=1)
 
-    data_dict = project.load_data.EinsteinDistinctDiffusionCoefficients()[0].data_dict
+    data_dict = project.load.EinsteinDistinctDiffusionCoefficients()[0].data_dict
 
     data = Path(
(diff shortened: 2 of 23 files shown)


mdsuite/calculators/einstein_diffusion_coefficients.py
```diff
@@ -25,6 +25,7 @@
 from tqdm import tqdm
 import tensorflow as tf
 from mdsuite.calculators.calculator import Calculator, call
```
SamTov, line 27: In the previous version did we not have some functionality for adding custom information to the table? Where is this now?

mdsuite/calculators/radial_distribution_function.py
```diff
@@ -355,13 +355,15 @@ def _calculate_radial_distribution_functions(self):
...
                 log.debug("Writing RDF to database!")
-                self._update_properties_file({
-                    "Property": "RDF",  # TODO this should be dynamic
-                    "Analysis": self.analysis_name,
-                    "subjects": names.split("_"),
-                    "data_range": self.data_range,
-                    "data": data
-                })
+
+                params = Parameters(
```
SamTov, line 359: For the prevention of updating all classes it looks like this can probably go into the parent class.

mdsuite/database/calculator_database.py
```diff
@@ -12,22 +12,32 @@
...
+from typing import TYPE_CHECKING, List, Union
 
 if TYPE_CHECKING:
     from mdsuite.experiment import Experiment
 
 log = logging.getLogger(__name__)
 
 
+@dataclass
+class Parameters:
```
SamTov, line 28: As data classes are not very intuitive nor are they really introduced to people it would be much better if you documented this appropriately in a doc string.

mdsuite/experiment/add_files.py
```diff
@@ -249,7 +249,7 @@ def build_species_dictionary(self):
         # self.save_class()
         self.species = species
 
-    def load_matrix(self, identifier: str = None, species: dict = None, select_slice: np.s_ = None, path: list = None):
+    def load_matrix(self, identifier: str = None, species: list = None, select_slice: np.s_ = None, path: list = None):
```
SamTov, line 252: We should discuss how all of this will work. It was initially intended that we remove this method all together as it was only called by experimental calculators or those that are not running correctly through our generators. In the case where we want users to be able to access the hdf5 data directly, some discussion should be had on the best way to do this. 

Specifically, for user interfacing we will need to go through the parameters and perhaps rename or simplify some.

mdsuite/calculators/einstein_diffusion_coefficients.py
PythonFZ: I assume  you are referring to `simulation_data` https://github.com/zincware/MDSuite-Core/blob/916a351aac91101878a412991a22378a7741e47d/mdsuite/database/experiment_database.py#L448-L535

mdsuite/calculators/radial_distribution_function.py
PythonFZ: Absolutely

mdsuite/database/calculator_database.py
PythonFZ: I might convert them to SQLAlchemy built-in mixins, but if this dataclass stays, I'll add a docstring

mdsuite/experiment/add_files.py
PythonFZ: We should probably do this on another PR, because this seems to be a bigger change including the data loading restructure

SamTov: Yes agreed. We can also wait for Tobias.

PR #331, opened 2021-09-20
Visualizer

### LINKED ISSUE #288: Issue while plotting RDF directly after POMF
**Describe the bug**
Because we use `fig, ax = plt.subplots()` and `plt.plot()` if you run e.g.,

```py
project.experiments["NaCl"].sample_rate = 1
# project.experiments["ml_md"].run_computation.RadialDistributionFunction(number_of_configurations=5000, minibatch=1, stop=5000)
project.experiments["NaCl"].run_computation.RadialDistributionFunction(
    number_of_configurations=10000, benchmark=True, minibatch=1, plot=True, batches=1, stop=10000
)
project.experiments["NaCl"].run_computation.PotentialOfMeanForce()
```

in a Jupyter notebook in a single cell, the output will look like:
image

**Potential fix**
Remove all `plt.plot()` references and purely use `fig, ax = plt.subplots()` to avoid issues like this.

SamTov:
Fix #288 

Implement a visualization module for MDSuite

* Add trajectory visualizer (Currently cannot step through time as we are waiting on the open3D group).
* Add a 3d visualizer for 3d calculators such as the spatial distribution function.
* Add 2d visualizer with bokeh and implement for all calculators. This now means we have access to full interactive plots in jupyter and in the browser. Bokeh also allows for the addition of powerful models and tools that we can add to any calculator.
* Added two example scripts, the larger of which will download data from the server. sdf.py does not and is here temporarily as I will continue to extend it.

Tests should suffice but ultimately we will need some plot based tests.


examples/scripts/green_kubo_diffusion.py
```diff
@@ -0,0 +1,89 @@
...
+    shutil.rmtree('GK_Diffusion_Example')
+
+
+if __name__ == '__main__':
+    """
+    Run the example.
+    """
+    load_data()  # load the data.
+    run_example()  # run the example.
+    clean_directory()  # clear the directory
```
PythonFZ: I'd prefer using `TemporaryDirectory`, it works on all systems and is designed for this use case.

```
import tempfile, os

temp_dir = tempfile.TemporaryDirectory()
os.chdir(temp_dir)
# run code
os.chdir("..")
temp_dir.cleanup()



```

examples/scripts/spatial_distribution_function.py
```diff
@@ -0,0 +1,35 @@
...
+    # #project.experiments.bmim_bf4.perform_transformation("UnwrapCoordinates")
+    # mapper = MolecularMap(
+    #     project.experiments.bmim_bf4,
+    #     molecules={
+    #         'bmim-7': {'smiles': 'CCCCN1C=C[N+](+C1)C', 'cutoff': 1.9},
+    #         'bf4-7': {'smiles': '[B-](F)(F)(F)F', 'cutoff': 2.4}
+    #               }
+    # )
+    # mapper.run_transformation()
+    # project.experiments.bmim_bf4.perform_transformation('WrapCoordinates')
```
PythonFZ: What about this part?

mdsuite/calculators/angular_distribution_function.py
```diff
@@ -421,7 +418,9 @@ def _compute_adfs(self, angles, species_indices):
         -------
         Updates the class, the SQL database, and plots values if required.
         """
-        for species in itertools.combinations_with_replacement(species_indices, 3):
+        for i, species in enumerate(itertools.combinations_with_replacement(species_indices, 3)):
+            if i == len(itertools.combinations_with_replacement(species_indices, 3)) - 1:
+                self.last_iteration = True
```
PythonFZ: ~~what is the purpose of `self.last_iteration` ?~~

Maybe rename it `computation_finished` which would make it a little more clear (to me).
Also does it have to be in the loop, or would it be possible to set the argument after the loop?

mdsuite/calculators/computations_dict.py
```diff
@@ -46,7 +46,7 @@
     'viscosity': {}
 }
 
-calculators = [EinsteinDiffusionCoefficients, EinsteinDistinctDiffusionCoefficients, GreenKuboSelfDiffusionCoefficients,
+calculators = [EinsteinDiffusionCoefficients, EinsteinDistinctDiffusionCoefficients, GreenKuboDiffusionCoefficients,
```
PythonFZ: I do think that `computations_dict` can be removed.

mdsuite/visualizer/d3_data_visualizer.py
```diff
@@ -0,0 +1,67 @@
+"""
+Visualizer for three dimensional data.
+"""
+import open3d as o3d
```
PythonFZ, line 4: I can neither `pip install open3d` nor `conda install -c open3d-admin -c conda-forge open3d` on windows.
Unfortunately, I think we have to make the 3D Visualizer an optional feature, keeping MDSuite compatible with all OS

mdsuite/calculators/coordination_number_calculation.py
```diff
@@ -242,31 +242,14 @@ def _get_coordination_numbers(self):
...
     def run_post_generation_analysis(self):
         """
         Calculate the coordination numbers and perform error analysis
         """
-
-        for data in self._get_rdf_data():  # Loop over all existing RDFs
+        calculations = self._get_rdf_data()
+        for i, data in enumerate(calculations):  # Loop over all existing RDFs
+            if i == len(calculations) - 1:
+                self.last_iteration = True
```
PythonFZ: We could do:

```
for i in x:
   self.draw(x_data, x_data, title, layouts)

self.show()
```

where

```
def show(self):
  self.plotter.grid_show(self.plot_array)
```

and remove the `last_iteration` entirely?

PythonFZ: + the labels on the plots are not displayed correctly

mdsuite/calculators/einstein_diffusion_coefficients.py
```diff
@@ -91,9 +90,9 @@ def __init__(self, **kwargs):
         self.species = None
         self.molecules = None
         self.database_group = 'Diffusion_Coefficients'
-        self.x_label = 'Time (s)'
-        self.y_label = 'MSD (m$^2$)'
-        self.analysis_name = 'Einstein_Self_Diffusion_Coefficients'
+        self.x_label = '$$Time (s)$$'
+        self.y_label = '$$MSD (m^{2})$$'
```
PythonFZ: I would change it to `Time / $$s$$` and `MSD / $$m^{2}$$` in accordance with SI unit system (https://www.nist.gov/pml/special-publication-811/nist-guide-si-chapter-7-rules-and-style-conventions-expressing-values)

If we want to keep `Time (s)` we still have to exclude the Time from the math mode

mdsuite/calculators/einstein_helfand_ionic_conductivity.py
```diff
@@ -72,9 +73,9 @@ def __init__(self, **kwargs):
         self.system_property = True
 
         self.database_group = 'Ionic_Conductivity'  # Which database_path group to save the tensor_values in
-        self.x_label = 'Time (s)'
-        self.y_label = 'MSD (m$^2$/s)'
-        self.analysis_name = 'Einstein_Helfand_Ionic_Conductivity'
+        self.x_label = '$$Time (s)$$'
+        self.y_label = '$$MSD (m^2/s)$$'
```
PythonFZ: See above, change format

mdsuite/graph_modules/molecular_graph.py
```diff
@@ -197,4 +197,6 @@ def reduce_graphs(adjacency_matrix: tf.Tensor):
         for item in del_list:
             molecules.pop(item)
 
+        print(len(molecules))
```
PythonFZ: we should use `log.<level>` or remove the print

mdsuite/utils/linalg.py
```diff
@@ -113,3 +113,70 @@ def apply_system_cutoff(tensor: tf.Tensor, cutoff: float) -> tf.Tensor:
...
+  """
+  with tf.name_scope(name):
+    point_spherical = tf.convert_to_tensor(value=point_spherical)
+
+    r, theta, phi = tf.unstack(point_spherical, axis=-1)
+    tmp = r * tf.sin(theta)
+    x = tmp * tf.cos(phi)
+    y = tmp * tf.sin(phi)
+    z = r * tf.cos(theta)
+    return tf.stack((x, y, z), axis=-1)
```
PythonFZ: flake8 missing new line

mdsuite/visualizer/__init__.py
```diff
@@ -2,8 +2,9 @@
...
 https://www.eclipse.org/legal/epl-v20.html
-
 SPDX-License-Identifier: EPL-2.0
 
-Copyright Contributors to the MDSuite Project.
-"""
+Copyright Contributors to the Zincware Project.
+
+Description: 
+"""
```
PythonFZ: flake8 missing new line

requirements.txt
```diff
@@ -21,4 +21,6 @@ sqlalchemy>=1.4
 pandas
 IPython
 pandoc
-tensorflow_probability
\ No newline at end of file
+tensorflow_probability
+open3d
```
PythonFZ, line 25: See issue about open3d above

PythonFZ (changes requested):
I like the new visualizer, looks very good!

But there are a few things that I think we should change:

- Do not open a new Tab when using jupyter
- The plots are too large inside notebooks
- The axis labels are not properly formatted
- remove `self.last_iteration` in favor of calling `self.show()` after the iteration is done. This looks more clean, reduces the amount of code and is easier to implement and maintain
- make `open3d` optional, because it can not be easily installed on Windows.

SamTov:
>   *<del> Do not open a new Tab when using jupyter </del>

This could be challenging unless we also have some jupyter notebook handling. I looked into how to work out if someone is using a notebook or not but it is quite difficult from what I understand. I can go and look into it more.

>    <del>  * The plots are too large inside notebooks </del>

This I can sort out I assume
 
>    <del>  * The axis labels are not properly formatted </del>

What do you mean by this? One issue with the labels is that the whole thing has to be under TeX in order for any TeX to render. If you are having issues with the rendering of the tex in the labels then it is likely a version issue as Bokeh only added next support 6 days ago. In this case I will need to add a version to the package.
 
>    * <del>  remove `self.last_iteration` in favor of calling `self.show()` after the iteration is done. This looks more clean, reduces the amount of code and is easier to implement and maintain </del>

I have a couple of ideas about the best way to handle the generation of thee plots. We could actually set up a html template and build a webpage essentially with the results. That way it will all be on one page.
 
>    * <del>  make `open3d` optional, because it can not be easily installed on Windows </del>

Damn... How hard is it? I will make it optional in the meantime while everyone replaces their computers with ones that work.

examples/scripts/spatial_distribution_function.py
SamTov: Yeah the example here is not complete yet. Before dev gets merged into main it will be. This will re-run the mol mapping which won't break anything but is time consuming.

examples/scripts/green_kubo_diffusion.py
SamTov: Okay I can add that in.

mdsuite/calculators/angular_distribution_function.py
SamTov: Certainly can be moved out of the loop. I would also like to think of some other formatting methods for these files as it is a lot of output and could be condensed.

mdsuite/calculators/computations_dict.py
SamTov: I will try and see what happens.

mdsuite/calculators/coordination_number_calculation.py
SamTov: Can you elaborate on not displayed correctly?

SamTov: The added minimum bokeh version should resolve issues with the rendering of LaTex. If it was something else please let me know.

mdsuite/visualizer/d3_data_visualizer.py
PythonFZ: Related to Python 3.9, using pre-builds http://www.open3d.org/docs/latest/getting_started.html#development-version-pip or downgrade to Python 3.8 solved the issue

mdsuite/calculators/coordination_number_calculation.py
PythonFZ: Yes, there was an issue with the math mode

examples/scripts/spatial_distribution_function.py
SamTov: I think I will extend this to also include a trajectory visualizer example so I will update this file today.

mdsuite/visualizer/d3_data_visualizer.py
SamTov: For now this can be resolved as we will not handle it at the present.

mdsuite/calculators/einstein_diffusion_coefficients.py
SamTov: Should now be resolved.

mdsuite/calculators/angular_distribution_function.py
SamTov: Should now be resolved.

mdsuite/calculators/calculator.py
```diff
@@ -778,6 +774,8 @@ def perform_computation(self):
 
             self._apply_averaging_factor()
             self._post_operation_processes()
+            if self.plot:
+                self.plotter.grid_show(self.plot_array)
```
PythonFZ: Maybe it got lost somewhere, but I think that this `if self.plot` can be moved outside of the outer if clause and can be put at the end

mdsuite/project/project.py
```diff
@@ -82,6 +83,11 @@ def __init__(self, name: str = None, storage_path: str = "./", description: str
             self.name = name
         self.storage_path = storage_path
 
+        if config.jupyter:
+            log.warning("Using Jupyter")
+        else:
+            log.warning("Not using Jupyter")
+
```
PythonFZ: Do we want to keep that?

PythonFZ (changes requested):

PythonFZ:
> Why can this one be moved? Theoretically we could move the actual plotters out of the if statement but if this is called but no plots generated it will load an empty screen

Because it is repeated 4 times in the lines 778, 787, 794, 832 and is always called last. You should be able to reduce it to a single statement after the `if self.system_property, elif, elfi, else`

This is now resolved

mdsuite/calculators/radial_distribution_function.py
```diff
@@ -95,8 +94,8 @@ def __init__(self, **kwargs):
         self.scale_function = {'quadratic': {'outer_scale_factor': 1}}
         self.loaded_property = 'Positions'
         self.database_group = 'Radial_Distribution_Function'  # Which database_path group to save the tensor_values in
-        self.x_label = r'r ($\AA$)'
-        self.y_label = 'g(r)'
+        self.x_label = r'$$\text{r} \AA$$'
```
PythonFZ: `\AA` does not work with Bokeh `'2.4.0'`

mdsuite/project/project.py
SamTov: Probably not

PythonFZ (approved):

mdsuite/calculators/radial_distribution_function.py
SamTov: has been removed from the rdf and the coordination numbers

PR #333, opened 2021-09-21
Add Tutorial 

### LINKED ISSUE #332: Add a tutorial data method.
We could have a data loading method for the tutorials and CI built into MDSuite which just makes a temp directory, downloads some data and changes into the temp directory for further operations.

It would make loading data really easy for our test cases and also fit nicely into the api.

Something like this currently exists in the green_kubo example found under MDSuite/examples/scripts

PythonFZ:
Fix #332

### DIFF
--- a/mdsuite/project/project.py
+++ b/mdsuite/project/project.py
@@ -29,6 +29,7 @@ from mdsuite.calculators import RunComputation
 from mdsuite.database.project_database import ProjectDatabase
 import mdsuite.database.scheme as db
 from dataclasses import make_dataclass
+from tempfile import TemporaryDirectory
 
 from typing import TYPE_CHECKING, Dict, List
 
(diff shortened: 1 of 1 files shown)


SamTov (commented):
This is excellent however I think we could have it be more general if it were its own module which just handled downloading different sets of data in the ExampleData repo? What do you think?

PythonFZ:
Maybe, but I don't know if would be worth writing a complex interface for a tutorial that will probably not be used by many. I think we should consider a tutorial situation where most if not all calculators and functions can be looked at and do not go beyond that.
If you have some ideas on how to make it look better, I will not argue against it.

SamTov:
With the new DataHub and ZincHub repositories, loading data for examples is now quite simple and the need for this PR is gone. I would like to close it @PythonFZ.

PythonFZ:
> With the new DataHub and ZincHub repositories, loading data for examples is now quite simple and the need for this PR is gone. I would like to close it @PythonFZ.

I also thinkt this should be closed and the examples provided will use ZnHub and make usinge example data sufficiently easy.

PR #337, opened 2021-09-23
Optimize and fix the SDF

PythonFZ:

* Speed up the `Building molecules`
```py
a = tf.range(100, 200)
b = tf.range(101)
check_a_in_b(a, b) == any(x in a for x in b)
check_a_in_b(b, a) == any(x in b for x in a)
# 18 ms vs 1.54 s
```
* fix the SDF slicing so that only the distances `species[0] - species[1]` are considered.
* Apply black coding style


SamTov (approved):
The code looks great. I can't see anything problematic.

PR #349, opened 2021-10-01
optimized calculator base class

PythonFZ:
- Clean up the calculator `__call__` method
- Add docstrings
- Add the new method to the RDF as an Example


SamTov:
This is now contained within #352

PR #352, opened 2021-10-03
Sam tov update generators

### LINKED ISSUE #275: Inconsistent arguments
**Describe the bug**
Some functions, like the ADF use `n_confs` as a paramter, others like the RDF use `number_of_configurations`.

We should make sure, that the arguments are unified.

### LINKED ISSUE #287: Reduction of the calculator class.
**Describe the bug**
The calculator class should probably be reduced in size. Many of the methods apply only to some of the calculators. I would suggest we discuss what kinds of parent classes we could use for some of the calculators as they do generally take on some common form. In this case, the calculator class would be quite minimal and act as a good template for custom calculators.

### LINKED ISSUE #289: Display the PMF in units of eV instead of J
The PMF values are given in Joule (~ e-19) and it might be more readable and compareable if we gave them in eV.
We could make this an argument, so one could select eV or J and this would then also be saved in the database.
In general it might make sense to store units in the database.

### LINKED ISSUE #297: Add self.dtype to calculators
We should make the type of the data in a calculator an attribute of the class. In this way, it is easier to track the data types when performing operation and requires only a single change. This should be done for all calculators. It will also allow us to set a standard for data types.

### LINKED ISSUE #324: experiment.load_matrix signature
species is speciefied as dict, but should be a list.
also: experiment.load_data could be called load_computation (or just load, then run_computation -> run)

### LINKED ISSUE #325: experiment.load_matrix return
It would be nice if the type was always the same and the shape was documented

### LINKED ISSUE #334: Inconsistent units with the RDF/ADF/SDF calculators

### LINKED ISSUE #366: self.update_user_args should not set the database

SamTov:
* Fixes #325 
* Fixes #324 
* Fixes #287 
* Fixes #275
* Fixes #297 
* Fixes #289
* Fixes #334 

TBD

- calling `project.add_experiment(data=...)` without an experiment name now raises an Error. Either pass `None` to autogenerate a Name or use a custom name. This change was introduced, because running the same code twice without a name will generate a new experiment every time. To avoid this the default has been changed.


PythonFZ (approved):
Due to the large size of the PR and my personal contribution to it I will approve this PR.
I updated the tests and looked at the docs. All looks good. So it should be good to merge.

PR #357, opened 2021-10-04
Improve db

PythonFZ:
Improve the `experiment_database.py` file by introducing a simple common method and moving everything to json entries in the database


SamTov (approved):
Seems okay but I cannot follow all the database changes. Pending passed tests I am okay.

PR #359, opened 2021-10-07
optimize the experiment database

PythonFZ:
- Adding new database entries is now simplified by calling `my_db_entry = LazyProperty()`
- DB entries are now cached in memory which can speed up all parts of the code that access them regularly, e.g. often query `temperature`


SamTov (approved):
looks nice

PR #360, opened 2021-10-19
Sam tov fix docs

SamTov:
Fixed the docs build

* Removed matplotlib extension to avoid the doc failure
* Found new failure with tf.Tensor type and opened an issue here `https://github.com/sphinx-doc/sphinx/issues/9754`

I will wait for this to be resolved before merging.


SamTov:
numpydoc fails for numpy doc strings but napolean does not. In our case, both were imported. This is now resolved. The issues that persist require renaming in the SQL database class or are related to doc features we don't want to display yet as they are out of date.

PythonFZ (approved):

PythonFZ:
tools.rst
```py
import mdsuite as mds
NaCl = mds.Experiment(...)
NaCl.unwrap_coordinates()
```

`NaCl.write_xyz(property='Unwrapped_Positions')` property is a built-in and should be renamed

PR #361, opened 2021-10-22
update get dimensionality method.

SamTov:

* Optimised the get dimensionality method.


mdsuite/utils/meta_functions.py
```diff
@@ -89,12 +88,15 @@ def join_path(a, b):
 
 def get_dimensionality(box: list) -> int:
     """
-    Calculate the dimensionality of the experiment box
+    Calculate the dimensionality of the experiment box.
```
SamTov, line 98: Did you write a test?

SamTov (commented):
Please address comments.

mdsuite/utils/meta_functions.py
SamTov: no,

mdsuite/utils/meta_functions.py
```diff
@@ -103,25 +105,7 @@ def get_dimensionality(box: list) -> int:
...
-            dimensions = 1
-        else:
-            dimensions = 2
-
-    # Other option is 3 dimensions.
-    else:
-        dimensions = 3
-
-    return dimensions
+    return np.where(box != 0, 0, 1, 0).sum()
```
PythonFZ, line 115: Please add a test

PythonFZ (changes requested):
Test missing

mdsuite/utils/meta_functions.py
SamTov: It seems thre has actually been a test for this the whole time. One of the first added.

```python
def test_get_dimensionality(self):
        """
        Test the get_dimensionality method.
        Returns
        -------
        assert that for all choices of dimension array the correct dimension comes out.
        """
        one_d = [1, 0, 0]
        two_d = [1, 1, 0]
        three_d = [1, 1, 1]
        assert get_dimensionality(one_d) == 1
        assert get_dimensionality(two_d) == 2
        assert get_dimensionality(three_d) == 3
```

PR #362, opened 2021-10-25
Fix broken tests

PythonFZ:
The tests that rely on the example data are broken, because the repository was renamed.
This PR introduces the `zinchub` package and attempts to fix the tests.


SamTov (approved):
Looks like a god update.

PR #365, opened 2021-10-27
typo...

PythonFZ:
~~this is why we maybe should be able to push to dev directly?~~


SamTov:
I don't think a typo would warrant an emergency push...

SamTov (approved):
Great catch.

PR #368, opened 2021-11-02
added versioneer capability

Fratorhe:
I implemented the versioneer system to keep track of versions more easily. 
This allows to assign the package version only on the git tags and the package will get it automatically. 
It also tracks versions with commits using the hash. 
I find it very useful to avoid writing it in several places and be more consistent. 
In addition, it makes it easier for error reporting since users can specify the exact version easily. 

* Added files to perform version check . 

```
pip install . 
import mdsuite
print(mdsuite.__version__)
```


SamTov:
I'm not a huge fan of how you have to copy a bunch of source code over into the project. There are a few projects that manage versioning albeit none quite as extensively as this one. But I am not sure if this comes at the cost of a nicely implemented tool.

SamTov:
Also, have we agreed on using the semantic versioning criteria? I think it makes the most sense and seems as though that is what we are using.

Fratorhe:
I see your point of adding the source code, but I guess it is to avoid dependencies. 
The code is in CC-0 which means there is no copyright, so it should be ok. 
There is also a lite version of this from the same author: https://github.com/jbweston/miniver

From my experience, most software uses semantic versioning and it is the one that seems most logic to me, so I'd go for that. 

I implemented this to avoid having to change the version number in several places. I am quite forgetful, and if something can be automatized and only be changed in one place, I will go for it.

SamTov:
There are certainly benefits to having the versioning accessible from within the code. I might just take a bit to think over if any improvement could be made in the following days.

PR #372, opened 2021-11-03
WIP: Refactor fileread

### LINKED ISSUE #302: add_data does not support `pathlib.Path`
**Describe the bug**

You can not pass a `pathlib.Path` object to the add_data method.
It does only work with strings.

Put a check `isinstance` together with `Path.as_posix()` in place, to allow working with pathlib objects

### LINKED ISSUE #312: Make unit tests easy

### LINKED ISSUE #367: Resolve added file path when checking read_files
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**System parameters**
Please run the following in your python kernel and copy the output below.

```python
import mdsuite as mds
mds.Report()
```
 **Output here**

**Additional context**

christophlohrmann:
separate filereading from the experiment class
make it easier to write new file readers

Todo

- [x] write new filereader baseclass
- [x] rewrite experiment class to use the new metadata format
- [x] adapt all file readers
- [x] test
- [ ] write online doc

Feedback is welcome at any stage


christophlohrmann:
status at the moment:
new numpy file reader and lammps seem to work.
calculators and transformators run, but produce wrong output, which I have a hard time tracking down

christophlohrmann:
If the flux file raders and extxyz will be replaced soon, I am not too motivated to get them working aswell.
If we treat a flux as a species with 1 particle, there is no change on the experiment/database side in terms of data handling

CI/unit_tests/database/test_experiment_database.py
```diff
@@ -47,7 +49,11 @@ def test_read_files(tmp_path, traj_file):
     """Test that read_files is saved correctly"""
     os.chdir(tmp_path)
     project_1 = mds.Project()
-    project_1.add_experiment(experiment="Exp01", data=traj_file)
+    file_proc = mdsuite.file_io.lammps_trajectory_files.LAMMPSTrajectoryFile(traj_file)
+    project_1.add_experiment(
+        experiment="Exp01", timestep=0.1
+    )  # todo bad place for time step
+    project_1.experiments["Exp01"].add_data(file_proc)
```
PythonFZ: I agree, this should be changed

CI/unit_tests/database/test_file_read.py
```diff
@@ -0,0 +1,140 @@
...
+        box_l=3 * [1.1],
+    )
+    data = TrajectoryChunkData(species_list=species_list, chunk_size=n_configs)
+    data.add_data(positions, 0, sp_name, "Positions")
+    data.add_data(velocities, 0, sp_name, "Velocities")
+
+    proc = script_input.ScriptInput(data=data, metadata=metadata, name="test_name")
+
+    with tempfile.TemporaryDirectory() as temp_dir:
+        project = mdsuite.Project(name="test_proj", storage_path=temp_dir)
```
PythonFZ: pytest provides a ficture for handling temporary directories automatically.
```
def test_build_dataset(tmp_path):
    os.chdir(tmp_path)
    # and / or
    project = mdsuite.Project(name="test_proj", storage_path=tmp_path)
  
```

mdsuite/database/simulation_database.py
```diff
@@ -56,6 +58,156 @@
...
+
+    Attributes
+    ----------
+    name
+        Name of the species (e.g. 'Na')
+    n_particles
+        Number of particles of that species
+    properties: list of PropertyInfo
+        List of the properties that were recorded for the species
+    mass and charge are optional
```
PythonFZ, line 74: Why are mass and charge not defined via `PropertyInfo` ?

mdsuite/database/simulation_database.py
```diff
@@ -197,137 +349,50 @@ def open(self, mode: str = "a") -> hf.File:
...
+                for prop_info in sp_info.properties:
+                    dataset_name = f"{sp_info.name}/{prop_info.name}"
+                    write_data = chunk_data[sp_info.name][prop_info.name]
+
+                    dataset_shape = database[dataset_name].shape
+                    if len(dataset_shape) == 2:
+                        # only one particle
+                        database[dataset_name][start_idx:stop_index, :] = np.squeeze(
+                            write_data[:, 0, :]
+                        )
```
PythonFZ: Is the `np.squeeze` safe here? It might be possible that it reduced the dimensionality to far?

mdsuite/file_io/script_input.py
```diff
@@ -0,0 +1,42 @@
+import mdsuite.file_io.file_read
+import typing
+
+from mdsuite.database.simulation_database import TrajectoryChunkData, TrajectoryMetadata
+
+
+class ScriptInput(mdsuite.file_io.file_read.FileProcessor):
+    """
+    For testing purposes. Does not actually process files, instead uses data given on instantiation
+    """
```
PythonFZ, line 10: Maybe we should collect such classes needed for testing somewhere.
Or have a naming convention that they can be easily identified?

mdsuite/transformations/transformations.py
```diff
@@ -267,30 +271,50 @@ def _save_coordinates(
...
+        # data_structure is dict {'/path/to/property':{'indices':irrelevant, 'columns':deduce->deduce n_dims, 'length':n_particles}
+        species_list = list()
+        # data structure only has 1 element
+        key, val = list(data_structure.items())[0]
+        path = str(copy.copy(key))
+        path.rstrip("/")
+        path = path.split("/")
+        prop_name = path[-1]
+        sp_name = path[-2]
+        n_particles = val.get("length", 1)
```
PythonFZ: this part is a little hard to read altough not difficult. Does it make sense to put it in a function?

mdsuite/transformations/transformations.py
```diff
@@ -267,30 +271,50 @@ def _save_coordinates(
...
+        )
+
+        try:
+            self.database.add_data(chunk=chunk, start_idx=index + self.offset)
         except OSError:
             """
             This is used because in Windows and in WSL we got the error that
             the file was still open while it should already be closed. So, we
             wait, and we add again.
             """
```
PythonFZ, line 320: I don't know but I have the feeling that `time.sleep(0.5)` was important here.

PythonFZ (commented):
I added some comments on parts of the code that I thought might be important to discuss.
Also a lot of open todos. But in general I really like how it looks and that adding data now is much cleaner than before! Great job.

mdsuite/transformations/transformations.py
Fratorhe: I think I added that...
I had some issues with file handling in Windows and adding this small delay fixed it. 
We can try without it again, maybe with the refactoring it is not needed anymore

christophlohrmann: woops sorry I did not mean to remove it

mdsuite/transformations/transformations.py
christophlohrmann: this is an ugly workaround that I intend to only keep until the transformation refactor. Once transformations use TrajectoryChunkData, this becomes obsolete

mdsuite/file_io/script_input.py
christophlohrmann: I fully intend to abuse this class for my pandas files created from espresso. So I like that it sits with the other filereaders

mdsuite/database/simulation_database.py
christophlohrmann: good point

mdsuite/database/simulation_database.py
christophlohrmann: I thought of a property as something that changes along the time, and other species properties as being fixed for the species, I did not anticipate that charge can also be time-dependent

CI/unit_tests/database/test_experiment_database.py
christophlohrmann: the todos are here to keep the PR somewhat self contained. The transformators and 'system property' are in for a checkup

mdsuite/database/simulation_database.py
PythonFZ: I guess it makes sense then to leave them here as a special type?

christophlohrmann: I would say so. The calculators should look for per-time-step values and if they are not there, fall back to the species default. Especially for mass I would say most of the time it is constant for all particles for all time steps

SamTov:
> If the flux file raders and extxyz will be replaced soon, I am not too motivated to get them working aswell.
> If we treat a flux as a species with 1 particle, there is no change on the experiment/database side in terms of data handling

I do not think he flux files or extxyz will be changes "soon". The chemfiles library does not handle flux files at all and its handling of text files is where we would need to loop over atoms so I think these readers will stay constant for the time being. I can also look at them though as the restructure continues. Great idea about the flux files, the only thing is that they do not have repeated headers much like in your numpy files.

mdsuite/database/simulation_database.py
SamTov: charge can be defined within the time-dependent information so I would be careful wth that one. Mass should remain constant.

mdsuite/experiment/experiment.py
```diff
@@ -396,3 +410,215 @@ def set_mass(self, element: str, mass: float):
...
+            )
+            return  # End the method.
+
+        database = Database(
+            name=pathlib.Path(self.database_path, "database.hdf5").as_posix()
+        )
+
+        # Check to see if a database_path exists
+        # todo make method of database class
+        database_path = pathlib.Path(
```
SamTov: I guess a `.exists` attribute could be the easiest way.

mdsuite/experiment/experiment.py
```diff
@@ -396,3 +410,215 @@ def set_mass(self, element: str, mass: float):
...
+    except KeyError:
+        raise ValueError(
+            f"{file_format} not found! \n"
+            f"Available io formats are are: {[key for key in dict_file_io.keys()]}"
+        )
+    return class_file_io(file_path=trajectory_file, sort=sort), file_type
+
+
+def _species_list_to_architecture_dict(species_list, n_configurations):
+    # TODO let the database handler use the species list directly instead of the dict
```
SamTov, line 677: I like this idea

SamTov (commented):
Aside from the other comments I just made a few more. The important one though is related to the charges being read in from the simulation data as that is very relevant. 

The other thing I wanted to as in general is did you keep the methods for parsing custom dump data? Sometimes user pass something like:

```
c_idx[0] c_idx[1] c_idx[2]
```

and we can map say `Atomic_Stress` to `c_idx` and in this way store the custom keyword with the data correctly in the database. This is quite important for viscosity calculations and the like so should be kept in.

christophlohrmann:
> > If the flux file raders and extxyz will be replaced soon, I am not too motivated to get them working aswell.
> > If we treat a flux as a species with 1 particle, there is no change on the experiment/database side in terms of data handling
> 
> I do not think he flux files or extxyz will be changes "soon". The chemfiles library does not handle flux files at all and its handling of text files is where we would need to loop over atoms so I think these readers will stay constant for the time being. I can also look at them though as the restructure continues. Great idea about the flux files, the only thing is that they do not have repeated headers much like in your numpy files.

The PR cannot be merged before extxyz and flux are in. If needed I can take a look at them and get them up to speed. Do we have any tests that use these files?
I would say that the flux files should remain their own reader, just the return structure ([n_steps, 1, n_dims] for each property) should be the same as for the other readers. I do not think it would hurt to have the extra dimension of length 1 in the database. With the added benefit that all data to and from the simulation database is a three-tensor

christophlohrmann:
> Aside from the other comments I just made a few more. The important one though is related to the charges being read in from the simulation data as that is very relevant.
> 
> The other thing I wanted to as in general is did you keep the methods for parsing custom dump data? Sometimes user pass something like:
> 
> ```
> c_idx[0] c_idx[1] c_idx[2]
> ```
> 
> and we can map say `Atomic_Stress` to `c_idx` and in this way store the custom keyword with the data correctly in the database. This is quite important for viscosity calculations and the like so should be kept in.

If there is a charge in the file, it is read in at every timestep and stored in the simulation database. I just had in mind the case where a simulation program never changes the charge and thus does not write it to a trajectory file. In that case the user needs a place to put the constant charge.

About the ``c_idx``: I have to admit I have no idea what you are talking about :)

mdsuite/experiment/experiment.py
christophlohrmann: this is just a question of how big the PR will be. Because if I replace the species dict with the species list, I also need to change all data loading and find a way to store the species_list in the sql database which I do not have experience with

SamTov:
> > > If the flux file raders and extxyz will be replaced soon, I am not too motivated to get them working aswell.
> > > If we treat a flux as a species with 1 particle, there is no change on the experiment/database side in terms of data handling
> > 
> > 
> > I do not think he flux files or extxyz will be changes "soon". The chemfiles library does not handle flux files at all and its handling of text files is where we would need to loop over atoms so I think these readers will stay constant for the time being. I can also look at them though as the restructure continues. Great idea about the flux files, the only thing is that they do not have repeated headers much like in your numpy files.
> 
> The PR cannot be merged before extxyz and flux are in. If needed I can take a look at them and get them up to speed. Do we have any tests that use these files? I would say that the flux files should remain their own reader, just the return structure ([n_steps, 1, n_dims] for each property) should be the same as for the other readers. I do not think it would hurt to have the extra dimension of length 1 in the database. With the added benefit that all data to and from the simulation database is a three-tensor

That would simplify things greatly. @Fratorhe could you please weigh in.

mdsuite/experiment/experiment.py
SamTov: We can leave it for now and make an issue to be resolved after.

SamTov:
> > Aside from the other comments I just made a few more. The important one though is related to the charges being read in from the simulation data as that is very relevant.
> > The other thing I wanted to as in general is did you keep the methods for parsing custom dump data? Sometimes user pass something like:
> > ```
> > c_idx[0] c_idx[1] c_idx[2]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and we can map say `Atomic_Stress` to `c_idx` and in this way store the custom keyword with the data correctly in the database. This is quite important for viscosity calculations and the like so should be kept in.
> 
> If there is a charge in the file, it is read in at every timestep and stored in the simulation database. I just had in mind the case where a simulation program never changes the charge and thus does not write it to a trajectory file. In that case the user needs a place to put the constant charge.
> 
> About the `c_idx`: I have to admit I have no idea what you are talking about :)

I absolutely agree about the charge. We do also need the case for constant charge. Just making sure nothing is lost along the way during the restructure. 

As for custom file dumps:

```python
def add_data(
        self,
        trajectory_file: str = None,
        file_format: str = "lammps_traj",
        rename_cols: dict = None,
        sort: bool = False,
        force: bool = False,
    ):
        """
        Add tensor_values to the database_path
        Parameters
        ----------
        file_format :
                Format of the file being read in. Default is file_path
        trajectory_file : str
                Trajectory file to be process and added to the database_path.
        rename_cols : dict
                If this argument is given, the columns with names in the keys of the
                dictionary will be replaced with
                the values.
        sort : bool
                If true, the tensor_values will be sorted when being entered into the
                database_path.
        force : bool
                If true, a file will be read regardless of if it has already
                been seen.
        """
```

The rename cols argument here allows one to adjust the naming convention on a column. We set some of these, e.g.

```python
{"Positions": ['x', 'y', 'z']}
```

but in principle also allow someone to compute something special in lammps, for example the reduced momentum on the particle, dump it in their trajectory file as rp_x, rp_y, rp_z and then we read it when they give us:

```python
rename_cols={"Reduced_Momentum": ["rp_x", "rp_y", "rp_z"]}
```

This is very important for dumps that need the force acting on the particle or the atomic energy at one point in time which we cannot re-compute with potential information.

christophlohrmann:
> > > Aside from the other comments I just made a few more. The important one though is related to the charges being read in from the simulation data as that is very relevant.
> > > The other thing I wanted to as in general is did you keep the methods for parsing custom dump data? Sometimes user pass something like:
> > > ```
> > > c_idx[0] c_idx[1] c_idx[2]
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > and we can map say `Atomic_Stress` to `c_idx` and in this way store the custom keyword with the data correctly in the database. This is quite important for viscosity calculations and the like so should be kept in.
> > 
> > 
> > If there is a charge in the file, it is read in at every timestep and stored in the simulation database. I just had in mind the case where a simulation program never changes the charge and thus does not write it to a trajectory file. In that case the user needs a place to put the constant charge.
> > About the `c_idx`: I have to admit I have no idea what you are talking about :)
> 
> I absolutely agree about the charge. We do also need the case for constant charge. Just making sure nothing is lost along the way during the restructure.
> 
> As for custom file dumps:
> 
> ```python
> def add_data(
>         self,
>         trajectory_file: str = None,
>         file_format: str = "lammps_traj",
>         rename_cols: dict = None,
>         sort: bool = False,
>         force: bool = False,
>     ):
>         """
>         Add tensor_values to the database_path
>         Parameters
>         ----------
>         file_format :
>                 Format of the file being read in. Default is file_path
>         trajectory_file : str
>                 Trajectory file to be process and added to the database_path.
>         rename_cols : dict
>                 If this argument is given, the columns with names in the keys of the
>                 dictionary will be replaced with
>                 the values.
>         sort : bool
>                 If true, the tensor_values will be sorted when being entered into the
>                 database_path.
>         force : bool
>                 If true, a file will be read regardless of if it has already
>                 been seen.
>         """
> ```
> 
> The rename cols argument here allows one to adjust the naming convention on a column. We set some of these, e.g.
> 
> ```python
> {"Positions": ['x', 'y', 'z']}
> ```
> 
> but in principle also allow someone to compute something special in lammps, for example the reduced momentum on the particle, dump it in their trajectory file as rp_x, rp_y, rp_z and then we read it when they give us:
> 
> ```python
> rename_cols={"Reduced_Momentum": ["rp_x", "rp_y", "rp_z"]}
> ```
> 
> This is very important for dumps that need the force acting on the particle or the atomic energy at one point in time which we cannot re-compute with potential information.

Ok got it. I indeed did not keep that mechanism, but will bring it back.

Fratorhe:
> > > > If the flux file raders and extxyz will be replaced soon, I am not too motivated to get them working aswell.
> > > > If we treat a flux as a species with 1 particle, there is no change on the experiment/database side in terms of data handling
> > > 
> > > 
> > > I do not think he flux files or extxyz will be changes "soon". The chemfiles library does not handle flux files at all and its handling of text files is where we would need to loop over atoms so I think these readers will stay constant for the time being. I can also look at them though as the restructure continues. Great idea about the flux files, the only thing is that they do not have repeated headers much like in your numpy files.
> > 
> > 
> > The PR cannot be merged before extxyz and flux are in. If needed I can take a look at them and get them up to speed. Do we have any tests that use these files? I would say that the flux files should remain their own reader, just the return structure ([n_steps, 1, n_dims] for each property) should be the same as for the other readers. I do not think it would hurt to have the extra dimension of length 1 in the database. With the added benefit that all data to and from the simulation database is a three-tensor
> 
> That would simplify things greatly. @Fratorhe could you please weigh in.

Hi, I already had the idea to rework how the flux files are read. 
At least for LAMMPS, I generate these files using the `fix print_data`. 
However, this fix does not print the column names or other system properties, so the user has to add the header manually.
I think it would be more convenient to have a (json) file with the system parameters which gets read together with the flux file, so it does not need to be modified. 
I would need to go through the new implementation that @christophlohrmann  has done, but I can take care of that.

christophlohrmann:
> Hi, I already had the idea to rework how the flux files are read. At least for LAMMPS, I generate these files using the `fix print_data`. However, this fix does not print the column names or other system properties, so the user has to add the header manually. I think it would be more convenient to have a (json) file with the system parameters which gets read together with the flux file, so it does not need to be modified. I would need to go through the new implementation that @christophlohrmann has done, but I can take care of that.

Sounds good. One of the goals of the file reader refactor was to be able to pass multiple files, such that for example the metadata can be read from one and then used on another one. So what you have in mind should work.
I never used lammps so I do not know how general the format of such a json parameter file is and how well it translates to all lammps users

Fratorhe:
> > Hi, I already had the idea to rework how the flux files are read. At least for LAMMPS, I generate these files using the `fix print_data`. However, this fix does not print the column names or other system properties, so the user has to add the header manually. I think it would be more convenient to have a (json) file with the system parameters which gets read together with the flux file, so it does not need to be modified. I would need to go through the new implementation that @christophlohrmann has done, but I can take care of that.
> 
> Sounds good. One of the goals of the file reader refactor was to be able to pass multiple files, such that for example the metadata can be read from one and then used on another one. So what you have in mind should work. I never used lammps so I do not know how general the format of such a json parameter file is and how well it translates to all lammps users

Basically, this json parameter file would be something we would "request". Some properties are accessible from the LAMMPS input file while many others are difficult to extract because there could be ambiguity since LAMMPS allows to run several processes in the same file. I think it makes more sense that the user fills up a simple file with the required parameters.

christophlohrmann:
@Fratorhe can you send me a flux file (and possibly the extra parameter file)? Then I can write the reader for it

Fratorhe:
> @Fratorhe can you send me a flux file (and possibly the extra parameter file)? Then I can write the reader for it

Hey sorry, I really wanted to have a look at this, but I needed to wrap up some other work before...
This is the kind of file to expect. 
However, the header is normally only the first line (# fix print...) while the rest I added it manually (including the line with the column names). 
So for this to work, this added header has to be always the same, and that's a pain. 
Also you need to append these lines on the file, modifying the original one, which I do not really like neither.
I think it is better to have another file which contains that information and gets read along the flux file. 

[flux.txt](https://github.com/zincware/MDSuite/files/7528736/flux.txt)

christophlohrmann:
Done for now. I will add tests for the various file readers once the syntax for adding files to an experiment is figured out

- current status
  ``exp.add_data(file_name, file_format:str)`` will no longer work because different readers take different arguments (metadata that is not provided in the file itself, multiple files etc.)
- new syntax 1
  ``reader = readers.ExampleReader(file1, file2, **kwargs)``
    ``exp.add_data(reader)``
  pro: clean, obvious dependencies, disentangled, experiment calls the reader
  con: one extra line for the user, feels maybe a bit clunky for the simple readers that do in fact only take one file path as argument
- new syntax 2
  ``exp.add_data.ExampleReader(file1, file2, **kwargs)``
  pro: only one line, same structure as for the calculators, so very user friendly
  con: dependency unclear. reader is called by user. but the experiment is the one calling the actual data loading methods of the reader. This can surely be worked out though
  
any opinions, @SamTov @PythonFZ ?

christophlohrmann:
ping @SamTov @PythonFZ

SamTov:
> Done for now. I will add tests for the various file readers once the syntax for adding files to an experiment is figured out
> 
>     * current status
>       `exp.add_data(file_name, file_format:str)` will no longer work because different readers take different arguments (metadata that is not provided in the file itself, multiple files etc.)
> 
>     * new syntax 1
>       `reader = readers.ExampleReader(file1, file2, **kwargs)`
>       `exp.add_data(reader)`
>       pro: clean, obvious dependencies, disentangled, experiment calls the reader
>       con: one extra line for the user, feels maybe a bit clunky for the simple readers that do in fact only take one file path as argument
> 
>     * new syntax 2
>       `exp.add_data.ExampleReader(file1, file2, **kwargs)`
>       pro: only one line, same structure as for the calculators, so very user friendly
>       con: dependency unclear. reader is called by user. but the experiment is the one calling the actual data loading methods of the reader. This can surely be worked out though
> 
> 
> any opinions, @SamTov @PythonFZ ?

In terms of user friendly I like the second option. My idea would be to have an option basically. If a person has no custom types or dumps then there is no reason it cannot be automatic and try to solve the file type automatically. I would guess that 90% of people will just dump a lammps, gro, or whatever file in there and have it read with no issues. Also, in the case of binary, there are no custom dump options so the field of people using custom headers, whilst important, are probably small. 

I would say that we could try to have it such that if a person wants to pass a reader they can and that reader should be a class instance as written in approach 1. If they pass nothing the reader tries to read the file extension and work it out. If this reading fails, it throws an error and tells people how to load a reader. Then all we need to do is write up some documentation for file readers which I am happy to get started on.

christophlohrmann:
Data loading from lammps and extyxz tested.
@Fratorhe can you upload a generic flux file to DataHub so it can be used in automatic testing?

mdsuite/file_io/extxyz_files.py
```diff
@@ -0,0 +1,305 @@
...
+var_names = {
+    "Positions": "pos",
+    "Velocities": "vel",
+    "Forces": "force",
+    "Stress": "stress",
+    "PE": "energies",
+    "Time": "time",
+    "Lattice": "Lattice",
+    "Momenta": "momenta",
+}
```
PythonFZ: I think this could also be a dataclass? 
Even if not used as such and converted via `dataclasses.asdict` it looks cleaner

mdsuite/file_io/extxyz_files.py
```diff
@@ -0,0 +1,305 @@
...
+        self._mdata = None
+        self._properties_dict = None
+        self._species_dict = None
+        self._batch_size = None
+
+    def __str__(self):
+        return str(self.file_path)
+
+    def get_metadata(self):
+        with open(self.file_path, "r") as file:
```
PythonFZ: I leave the docstrings mostly to Sam, but this needs docstrings

mdsuite/file_io/extxyz_files.py
```diff
@@ -0,0 +1,305 @@
...
+            n_configurations=n_configs,
+            box_l=box_l,
+            sample_rate=sample_rate,
+            species_list=species_list,
+        )
+        self._batch_size = mdsuite.utils.meta_functions.optimize_batch_size(
+            filepath=self.file_path, number_of_configurations=n_configs
+        )
+
+        return self._mdata
```
PythonFZ: It feels like this should be a property. 
First because I like properties, but also because you use it internally as `self._mdata` and externally via `metadata = reader.get_metadata()` so why not combine both into a single `self.metadata` / `reader.metadata` and use 

```py
if self._metadata is None:
 self._metadata = self._get_metadata()
return self._metadata
```

mdsuite/file_io/extxyz_files.py
```diff
@@ -0,0 +1,305 @@
...
+        self.custom_data_map = custom_data_map
+
+        self._n_particles = None
+        self._mdata = None
+        self._properties_dict = None
+        self._species_dict = None
+        self._batch_size = None
+
+    def __str__(self):
+        return str(self.file_path)
```
PythonFZ: I think this can be in the parent class, because you use the same method in every file reader. If a filereader needs adjustments it can be overwritten

mdsuite/file_io/extxyz_files.py
```diff
@@ -0,0 +1,305 @@
...
+        )
+        self._batch_size = mdsuite.utils.meta_functions.optimize_batch_size(
+            filepath=self.file_path, number_of_configurations=n_configs
+        )
+
+        return self._mdata
+
+    def get_configurations_generator(
+        self,
+    ) -> typing.Iterator[mdsuite.file_io.file_read.TrajectoryChunkData]:
```
PythonFZ: docstrings + see #397 for the future

mdsuite/file_io/extxyz_files.py
```diff
@@ -0,0 +1,305 @@
...
+        for config_idx in range(n_configs):
+            # skip the header
+            mdsuite.file_io.file_read.skip_n_lines(file, self.n_header_lines)
+            # read one config
+            traj_data = np.stack(
+                [
+                    np.array(list(file.readline().split()))
+                    for _ in range(self._n_particles)
+                ]
+            )
```
PythonFZ: the inner `np.array` is probably not required. Speedwise there is no difference though 
image

mdsuite/file_io/lammps_trajectory_files.py
```diff
@@ -57,222 +62,247 @@
...
+            custom_data_map = {}
+        self.custom_data_map = custom_data_map
+
+        # attributes that will be filled in by get_metadata() and are later used by get_configurations_generator()
+        self._batch_size = None
+        self._id_column_idx = None
+        self._n_particles = None
+        self._species_dict = None
+        self._property_dict = None
+        self._mdata = None
```
PythonFZ: they seem to be similar across FileProcessors. Would it make sense to put them in the parent class?

mdsuite/file_io/lammps_trajectory_files.py
```diff
@@ -20,16 +20,21 @@
...
+import pathlib
+
+from mdsuite.database.simulation_database import (
+    TrajectoryMetadata,
+    PropertyInfo,
+    SpeciesInfo,
+)
 
 var_names = {
     "Positions": ["x", "y", "z"],
```
PythonFZ: This also screams dataclass to me.

mdsuite/file_io/lammps_trajectory_files.py
```diff
@@ -57,222 +62,247 @@
...
+        n_configs = self._mdata.n_configurations
+        n_batches, n_configs_remainder = divmod(int(n_configs), int(self._batch_size))
 
-        return int(number_of_lines / (number_of_atoms + self.header_lines))
+        with open(self.file_path, "r") as file:
+            file.seek(0)
+            for _ in tqdm.tqdm(range(n_batches)):
+                yield self._read_process_n_configurations(file, self._batch_size)
+            if n_configs_remainder > 0:
+                yield self._read_process_n_configurations(file, n_configs_remainder)
```
PythonFZ: parts of this are also either identical duplicates or only differ by 

```py
# skip the header
mdsuite.file_io.file_read.skip_n_lines(file, self.n_header_lines)
```

and could also be combined in the common parent class.

mdsuite/file_io/lammps_trajectory_files.py
```diff
@@ -57,222 +62,247 @@
...
+            for sp_info in self._mdata.species_list:
+                idxs = self._species_dict[sp_info.name]["line_idxs"]
+                sp_data = traj_data[idxs, :]
+                # slice by property
+                for prop_info in sp_info.properties:
+                    prop_column_idxs = self._property_dict[prop_info.name]
+                    write_data = sp_data[:, prop_column_idxs]
+                    # add 'time' axis. we only have one configuration to write
+                    write_data = write_data[np.newaxis, :, :]
+                    chunk.add_data(write_data, config_idx, sp_info.name, prop_info.name)
```
PythonFZ: this also seems to be a common method, which differes mostly by where/if to add a `np.newaxis`

mdsuite/project/project.py
```diff
@@ -130,16 +132,18 @@ def add_experiment(
         units: Union[str, Units] = None,
         cluster_mode: bool = None,
         active: bool = True,
-        data: Union[str, list, dict] = None,
+        fname_or_file_processor: Union[
+            str, mdsuite.file_io.file_read.FileProcessor, list
+        ] = None,
```
PythonFZ: Does this support pathlib as well, see #302

PythonFZ (commented):
Looks good! I'm looking forward to merging this soon. I added a few small comments exclusively on possible duplicate code fragments and code enhancements. So feel free to adress them if you agree on them being helpful.

mdsuite/file_io/extxyz_files.py
christophlohrmann: will do. we can have this in a single place

mdsuite/file_io/extxyz_files.py
christophlohrmann: the script input reader does not have a file path.

mdsuite/file_io/lammps_trajectory_files.py
SamTov: For a text file yes, for binary it cannot be so generic but I guess binary files will be exclusively handled by chemfiles and not read so manually by us.

mdsuite/file_io/lammps_trajectory_files.py
christophlohrmann: with regards to all the parent class arguments: I agree that a "TabularTextFileProcessor" could unite a few of the methods common to extxyz and lammpstraj. I will sneak in an intermediate class (or do it with common free functions, I will see which is best)

mdsuite/project/project.py
christophlohrmann: probably at the moment it even does, but it is not stated clearly. I will include the explicit option

mdsuite/file_io/lammps_trajectory_files.py
christophlohrmann: I will try to streamline the tabular-text-style readers.
For binary readers, it might be beneficial to not batch by time step but maybe by species or property

mdsuite/file_io/extxyz_files.py
PythonFZ: maybe wrap it in a try..except then? It feels a little weird to have this duplicated in every class?

christophlohrmann: my point was to force the reader dev to think about what makes the file reader instance unique

PythonFZ: It is not used anywhere in the code, only for representation to the user I guess, so forcing the develeper to implent it is a good idea.

christophlohrmann: it is used in the 'files already read'-check of the experiment. It is now in the tabular text parent class

CI/integration_tests/calculators/__test_einstein_distinct_diffusion_coefficients.py
```diff
@@ -58,7 +58,9 @@ def test_eddc_project(traj_files, true_values, tmp_path):
     """Test the EinsteinDistinctDiffusionCoefficients called from the project class"""
     os.chdir(tmp_path)
     project = mds.Project()
-    project.add_experiment("NaCl", data=traj_files[0], timestep=0.002, temperature=1400)
+    project.add_experiment(
+        "NaCl", fname_or_file_processor=traj_files[0], timestep=0.002, temperature=1400
```
SamTov: I think this argument name needs changing. Perhaps `experiment_data`, `simulation_data` or something along these lines.

CI/integration_tests/calculators/__test_green_kubo_distinct_diffusion_coefficients.py
```diff
@@ -59,7 +59,9 @@ def test_project(traj_files, true_values, tmp_path):
     """
     os.chdir(tmp_path)
     project = mds.Project()
-    project.add_experiment("NaCl", data=traj_files[0], timestep=0.002, temperature=1400)
+    project.add_experiment(
+        "NaCl", fname_or_file_processor=traj_files[0], timestep=0.002, temperature=1400
```
SamTov: same comment

SamTov (commented):
I think the parameter name needs to be changed. Otherwise I will leave the rest of the review to Fabian. I noticed a lot of doc string missing in methods and classes which should be updated.

mdsuite/experiment/experiment.py
```diff
@@ -401,3 +448,234 @@ def set_mass(self, element: str, mass: float):
...
+                log.debug(temp[0].exact_mass)
+            except (ElementMassAssignedZero, IndexError):
+                sp_info.mass = 0.0
+                log.warning(
+                    f"WARNING element {sp_info.name} has been assigned mass=0.0"
+                )
+    return species_list
+
+
+def _species_list_to_architecture_dict(species_list, n_configurations):
```
PythonFZ, line 676: Maybe add a return statement here, is it `{species_name: {property: <property>}}` ?

mdsuite/file_io/extxyz_files.py
```diff
@@ -0,0 +1,256 @@
...
+
+        return mdata
+
+    def _get_species_information(self, file, species_idx: int, n_particles: int):
+        """
+        Get the initial species information
+
+        Parameters
+        ----------
+
```
PythonFZ: missing

mdsuite/file_io/extxyz_files.py
```diff
@@ -0,0 +1,256 @@
...
+    else:
+        raise RuntimeError("Could not find lattice size in file header")
+
+    if stop is not None:
+        lattice = data[start : stop + 1]
+        lattice[0] = lattice[0].split("=")[1].replace('"', "")
+        lattice[-1] = lattice[-1].replace('"', "")
+        lattice = np.array(lattice).astype(float)
+
+    return [lattice[0], lattice[4], lattice[8]]
```
PythonFZ: does this return the diagonal elements of the box vectors? maybe add a docstring?

mdsuite/file_io/extxyz_files.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+def _get_time(header: str):
+    data = copy.deepcopy(header).split()
```
PythonFZ: what is time? / docstring

mdsuite/project/project.py
```diff
@@ -124,22 +126,21 @@ def __str__(self):
 
     def add_experiment(
         self,
-        experiment: str = NoneType,
+        experiment: str = NoneType,  # todo: rename to sth like 'name'
```
PythonFZ: Yes! We should adress this either in an issue or here directly. I like `name` because it is clear that it is the name of the experiment

PythonFZ (commented):
Some small docstring requests. And we should rename `add_experiment(experiment)` to `add_experiment(name)`

PythonFZ (approved):
Looks good!

PR #380, opened 2021-11-25
adjust plotting to allow for custom plots.

SamTov:
Changes:

* Added running integral to the vacf computation
* Adjusted plotting and demonstrated how to a very customised plots.

Example plot for vacf:

image

I have removed the GK test for a couple of reasons.

1. I have changed what gets stored in the SQL database and therefore would need to change the data in the test
2. Before doing this I want to move over to proper testing data as the stuff we use now is quite unhelpful. 

Once merged lets merge dev into main and remove dev.


PythonFZ:
Can you open an issue for the disabled test with `p1-high`?

examples/scripts/green_kubo_diffusion.py
```diff
@@ -68,21 +68,21 @@ def run_example():
...
+    # )
+    # project.run.EinsteinHelfandIonicConductivity(
+    #     data_range=100, correlation_time=10, plot=True
+    # )
+    #
+    # project.run.RadialDistributionFunction(
+    #     number_of_configurations=100, start=0, stop=101, plot=True
+    # )
+    # project.run.GreenKuboIonicConductivity(data_range=100)
+    #
```
PythonFZ: I think we can remove this entire file in favor of a Notebook using DataHub inside the Docs?

PythonFZ (changes requested):
Looks all good.
The two things that I'd like to change are
- [x] fix the documentation
- [x] add an Issue for the removed test with `p1-high`

PythonFZ (approved):

PR #385, opened 2021-11-30
avoid bokeh having scroll bars

### LINKED ISSUE #375: Bokeh plots should be smaller.
The plots are to large in JupyterLab

The following is taken from a 1920x1080 screen. They should be smaller.
image

Fratorhe:
I am not sure if this also fixes #375 
However, every time I was visualizing with Bokeh, Chrome had vertical scrolling and it was quite annoying. 
Adding `sizing_mode='stretch_both'` adapts it to the screen size. 

Very small PR, but otherwise I will forget about it.

### DIFF
--- a/mdsuite/utils/config.py
+++ b/mdsuite/utils/config.py
@@ -49,3 +49,7 @@ try:
     config.jupyter = get_ipython().__class__.__name__ == "ZMQInteractiveShell"
 except NameError:
     pass
+
+if config.jupyter:
+    # change default in Jupyter Notebooks
+    config.bokeh_sizing_mode = None
--- a/mdsuite/utils/config.py
+++ b/mdsuite/utils/config.py
@@ -29,10 +29,18 @@ from dataclasses import dataclass
 
 @dataclass
 class Config:
-    """Collection of MDSuite configurations"""
+    """Collection of MDSuite configurations
+
+    Attributes
+    -----------
+    bokeh_sizing_mode: str
+        The way bokeh scales plots.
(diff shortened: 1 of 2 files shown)


PythonFZ:
For me, this makes plots actually bigger and not smaller.
And it does change the background grid from being quadratic to rectangular.

New:
image
vs old:
image

Fratorhe:
mm that's in Jupyter notebook?
For me in a py script and with Chrome in Windows: 

New:
image

Old: 
image
image

And zooming out does not solve it. 

I have tried the other options provided in sizing_mode (https://docs.bokeh.org/en/latest/docs/reference/models/layouts.html#bokeh.models.LayoutDOM.sizing_mode). 
Any stretch option works "well". 
The options with scale do nothing.

PythonFZ:
Yes, it is in jupyter notebook.

Fratorhe:
Do you mind trying with a script with the old and see if you also have this problem? 
Otherwise it may be my Google Chrome or something else...

SamTov:
I think for now as long as the full image is on the screen between Jupyter and browser we are okay. We need a seperate solution to the Jupyter issue. I haven not had issues with browser but also have not tried with chrome so good catch

PythonFZ (approved):
Adding it as a `config.bokeh_sizing_mode` makes if more future proof between jupyter and html.

SamTov:
> Adding it as a `config.bokeh_sizing_mode` makes if more future proof between jupyter and html.

We can also add this into the sizing in Jupyter maybe

PR #387, opened 2021-12-01
add updated notebook.

SamTov:
Added a notebook to the examples and to the documentation. Also fixed the axis label on green Kubo ionic conductivity


requirements.txt
```diff
@@ -16,4 +16,4 @@ tensorflow_probability
 open3d
 colorutils
 bokeh
-pandoc
\ No newline at end of file
+pandoc
```
PythonFZ: Pandoc is not required for installing MDSuite, is it?

PythonFZ (changes requested):
Why do we need the `examples/notebooks/NaCl` when we have the one inside the docs?

Otherwise it looks good. I would like to add somewhere how to access the data generated by the calculators.

requirements.txt
SamTov: > Why do we need the `examples/notebooks/NaCl` when we have the one inside the docs?
> 
> Otherwise it looks good. I would like to add somewhere how to access the data generated by the calculators.

Good question, I actually would like it to be the other way around. In the ZnRND docs I have in the makefile that it copies the notebooks into the docs for the build and them removes them after. 

They should be in the examples directory so people can open and read the notebooks executing them manually in the examples file rather than go into the docs to find them. I would like to have what I have in ZnRND set up here but then I need to sort the comment below.

I would do this for MDSuite but need to time to learn how to write .bat files so it doesn't fail on windows.

SamTov: I don't remember if panda is required, it was for something because we had to add it to a GitHub actions script with a sudo call.

PythonFZ: Le't see what happens without it in the tests

PythonFZ: Seems to work just fine

PythonFZ: So we could add
https://github.com/zincware/ZnRND/blob/39ec61251660dd3b524d3c05476a7e2c7de0a7e7/docs/Makefile#L11-L14
this and should be fine.

SamTov: yep that should work fine as long as the paths are correct.

PythonFZ (approved):

PythonFZ (changes requested):
must fix RTD before merge

PythonFZ (approved):

PR #388, opened 2021-12-01
Improve accessing data 

PythonFZ:
Prior to this one had to use
```py
data = project.experiments.NaCl_1100K.run.EinsteinDiffusionCoefficients(data_range=1000, correlation_time=1, gpu=True)
data.data_dict["Na"]
```

Now one can use simply

`data["Na"]`


SamTov (approved):
I like it.

PR #405, opened 2021-12-07
Add isort

PythonFZ:
# DO NOT MERGE UNTIL PYTEST PASSES!

The package https://github.com/pycqa/isort handles how imports should be handled.
I think this adds a nice touch together with black coding style.

Found on https://github.com/hgrecco/pint


.isort.cfg
```diff
@@ -0,0 +1,2 @@
+[settings]
+line_length=88
```
SamTov: GitHub is complaining about a newline.

SamTov (approved):
Checked the config files, not checking the adjusted imports...

SamTov (changes requested):
Pytest fails.

SamTov (approved):
pytest now passes.

PR #419, opened 2021-12-08
New Transformations API

PythonFZ:
Add 
- `project.experiments.NaCl.run.<Transformation>()`
- `project.experiments.NaCl.cls_transformation_run(mdsuite.transformations.<Transformation>())`
- `project.run.<Transformation>()`

Missing
- [x] Tests
- [ ] exception handling for `ValueError: Unable to create dataset (name already exists)`
- [x] updating calculators
- [ ] removing old method
- [ ] fix type hints in `run.py`


mdsuite/graph_modules/molecular_graph.py
```diff
@@ -72,7 +77,7 @@ def __init__(
         self.species = species
 
         self.database = Database(
-            name=os.path.join(self.experiment.database_path, "database.hdf5"),
+            name=(self.experiment.database_path / "database.hdf5").as_posix(),
```
SamTov, line 80: Is this a feature of Path objects?

mdsuite/transformations/unwrap_coordinates.py
```diff
@@ -254,5 +254,13 @@ def run_transformation(self):
...
         """
         for item in self.species:
-            self._unwrap_coordinates(item)  # run the transformation.
+            exists = self.database.check_existence(join_path(item, "Positions"))
+            # Check if the tensor_values has already been unwrapped
+            if exists:
+                print(
+                    f"Unwrapped positions exists for {item}, "
+                    "using the saved coordinates"
+                )
```
PythonFZ: should use logging

mdsuite/calculators/trajectory_calculator.py
```diff
@@ -169,12 +171,12 @@ def _string_to_function(argument):
             switcher = {**switcher_unwrapping, **switcher_transformations}
 
             choice = switcher.get(
-                argument, lambda: "Data not in database and can not be generated."
+                argument, lambda: "Data not in database and cannot be generated."
             )
             return choice
```
PythonFZ: Not related to the PR, because this is only formatting, but this confuses me. I think this should either raise and Error or use logging instead of what it is doing here.

mdsuite/graph_modules/molecular_graph.py
PythonFZ: Yes, unfortunately the Database class does not handle pathlib objects, so we have to convert it back to a posix path.
I just read https://towardsdatascience.com/why-you-should-start-using-pathlib-as-an-alternative-to-the-os-module-d9eccd994745 for a small overview.

mdsuite/calculators/trajectory_calculator.py
SamTov: Yes it can be changed. It was the default Python example of implementing a switch case and is very old in the code. There were more of them but now I think this may be the last one.

SamTov (approved):
I think everything looks nice. Once tests pass we can marge.

PR #425, opened 2021-12-08
Sam tov atomwise batching

### LINKED ISSUE #390: Atom-wise batching for dynamics is broken.
**Describe the bug**
The atom-wise batching for dynamic calculations is broken. 

**To Reproduce**
Run a large analysis on limited memory and it will fail.

Related to #383

### LINKED ISSUE #407: Fix kirkwood_buff_integrals
https://github.com/zincware/MDSuite/runs/4446800766?check_suite_focus=true

it seems like kirkwood_buff_integrals can fail occasionally and this  should not be possible.

might be related to #383

SamTov:
Fix #407
Fix #390 

* Fixed atom-wise batching 
* Added memory management tests

Run a calculator requiring atom-wise batching.

#### Notes
Not included in this PR is a proper test for the data manager and batching tests. As this will require a more substantial test involving a constructed database I am leaving it for a separate PR to come soon.


CI/unit_tests/memory_manager/test_memory_manager.py
```diff
@@ -0,0 +1,248 @@
...
+        """
+        Test the linear scale selection.
+
+        Notes
+        -----
+        Each test will check that the function parameters are return correctly and
+        that the function is called correctly and returns proper values.
+        """
+        # Test linear function
+        scale_function = {"linear": {"scale_factor": 2}}
```
PythonFZ, line 79: I think this could also be a dataclass, so we know what types are available, `linear, log-linear, ...`

CI/unit_tests/memory_manager/test_memory_manager.py
```diff
@@ -0,0 +1,248 @@
...
+        """
+        Test the log-linear scale selection.
+
+        Notes
+        -----
+        Each test will check that the function parameters are return correctly and
+        that the function is called correctly and returns proper values.
+        """
+        # Test log-linear function
+        scale_function = {"log-linear": {"scale_factor": 2}}
```
PythonFZ, line 96: For me to understand this correctly, we do have the type of the scaling, e.g. `log-linear` and then the `scaling_factor` is in addition to that so `f(x) = scaling * x` for linear

CI/unit_tests/memory_manager/test_memory_manager.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def test_get_optimal_batch_size(self):
+        """
+        Test the get_optimal_batch_size method.
+
+        Returns
+        -------
+        Test that this method returns the expected value. Currently this is just
+        the same value that is passed to it.
+        """
```
PythonFZ, line 199: will this change? Maybe add a TODO

PythonFZ (changes requested):
It all looks good. Maybe you can address the comments if you think it would make sense in the future? I don't think it should be part of this PR so I would suggest making an issue if it makes sense to you.

There is one thing that needs to be changed though in the ipynb
I think the argument names have changed in the most recent version after #372 
```py
project.add_experiment(
        experiment="NaCl_example_data",
        timestep=0.002,
        temperature=1400.0,
        units="metal",
        data="NaCl_gk_i_q.lammpstraj",
    )

```

CI/unit_tests/memory_manager/test_memory_manager.py
SamTov: yep great idea.

PythonFZ (approved):

CI/unit_tests/memory_manager/test_memory_manager.py
SamTov: The test is not where the issue is. It is in the actual memory management. It won't be changed until we have a better approach to the memory management.

CI/unit_tests/memory_manager/test_memory_manager.py
SamTov: so the memory goes through the function and then is scaled by this factor, for log-linear:

```
memory_used = scaling * memory * log(memory)
```

CI/unit_tests/memory_manager/test_memory_manager.py
SamTov: Added #430 to approach this.

Issue #392, opened 2021-12-02
experiment.species should be a dataclass


PythonFZ:
Currently `experiment.species` is a dictionary but it would be benefitial to make it a dataclass, so instead of `species['Na']['charge']` one can do `species['Na'].charge`. This would also set the available keys and make things easier for implementing.

christophlohrmann:
the dataclass could be very similar (or actually be) mdsuite.database.simulation_database.SpeciesInfo (from PR #372).
I do not know how to store that in the SQL, though

PythonFZ:
It would not be to difficult. We can adapt
https://github.com/zincware/MDSuite/blob/bc41b70194a8dc4c8af288ca120363e15d0580b9/mdsuite/database/experiment_database.py#L196-L242
to work with the dataclass.
We would have to convert it to a JSON serializeable dict and then we can pass it through
https://github.com/zincware/MDSuite/blob/bc41b70194a8dc4c8af288ca120363e15d0580b9/mdsuite/database/scheme.py#L184

PR #429, opened 2021-12-08
Add DotDict

PythonFZ:
Remove all `len(species[x]["indices"])` and introduce DotDict as precursor to #392 

This is one of the first steps to introduce #392 but can be merged prior to this because on it's own it is already an improvement.


SamTov (approved):
This is a really nice addition to the code.

PR #441, opened 2021-12-10
chemfiles reader

christophlohrmann:
implement reader that uses chemfiles to access the trajectories

P.S. why is this thing 60 commits long?


SamTov (approved):
I think everything looks great. Was there any test done for the .trr files or .trj?

christophlohrmann:
nop, it just uses xtc

christophlohrmann:
why do we use chemfiles instead of mdanalysis? While writing the tests mdanalysis was very helpful for reading

SamTov:
The reason was avoiding the other baggage that came along with it. MDAnalysis sort of teeters on the edge of also being an analysis tool for biochemistry and I didn't want to drag all of that along with it. We could go to using it, I would also say that we have some relationship with the core developer of chemfiles now and that could be an opportunity to work with them to build something inline with what we need more specifically.

PR #442, opened 2021-12-10
Add pre-commit hooks

PythonFZ:
This will introduce pre-commit hooks that can be used to ensure that the tests succeed.

Use:

```
pip install pre-commit
pre-commit install
```

and then `git commit` will automatically run isort/black/flake8 checks for you. If they fail you need to add the changed files again and try another commit.

- Changed line length from 88 to 90


SamTov (approved):
It looks nice. Lot's of file changes due to the black run but the configuration files seem correct and the tests pass.

Maybe we should add some developer documentation for the setup.

PR #443, opened 2021-12-13
move nblink

PythonFZ:
Move the `NaCl_walkthrough.nblink` into a `_notebooks` directory.

Checked via `make html`


SamTov (approved):
Excellent.

PR #446, opened 2021-12-15
change load_matrix() return and signature

### LINKED ISSUE #445: load_matrix() should return a matrix
eperiment.load_matrix() has intricate logic to handle different arguments, but is only ever used to return a matrix of values for a single property of a single species. This could be simplified

christophlohrmann:
Fix #445 

I remember complaining about this before and the issue is not there anymore. What was the argument against this form?

### DIFF
--- a/CI/unit_tests/database/test_file_read.py
+++ b/CI/unit_tests/database/test_file_read.py
@@ -116,16 +116,12 @@ def test_read_script_input(tmp_path):
     exp.add_data(proc)
 
     pos_loaded = np.swapaxes(
-        exp.load_matrix(species=[sp_name], property_name="Positions")[
-            "test_species/Positions"
-        ].numpy(),
+        exp.load_matrix(species_name=sp_name, property_name="Positions").numpy(),
         0,
         1,
     )
     vel_loaded = np.swapaxes(
-        exp.load_matrix(species=[sp_name], property_name="Velocities")[
-            "test_species/Velocities"
-        ].numpy(),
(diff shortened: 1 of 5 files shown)


SamTov:
I discussed in the issue but the method here is legacy or was supposed to be removed. 

We discussed making a seperate data loading class for user to access data from the database. We wanted it to return a generator of the data from the database. This is seperate from the ones used in the internal calculators as they have memory management integrated into it.

christophlohrmann:
would we need a big data loading class for the user?
Seems to me as the only thing that needs to change is database.load_data() to return a generator and not the full dataset.
experiment.load_data() is just a convenience-method to get the correct path to the correct file.

SamTov:
So the database method is there to be called by the generators. So it can return as much data as it needs to and we use a wrapper for that to tell it how much to load. I like this as it means that the database can just return data and you manage the generator.

We can make an experiment.load_data() method which uses the database method to create a generator and that I think would be a great idea. Right now the database only return a dict as it is the most general method for the other methods. So the experiment method would have to re-format that and return a tensor of values.

christophlohrmann:
I agree on separation of database and generator stuff. I will try to read through the code and understand how that works

SamTov:
Awesome. The only methods are really in the data_manager. There are perhaps some legacy things laying around there as they weren't removed due to tests not being comprehensive enough. At this stage the data managers will always return a dictionary.

SamTov:
@christophlohrmann Can this be closed?

Issue #383, opened 2021-11-25
Batching untested


christophlohrmann:
I am pretty sure batching is broken somewhere, because I get random test failures depending on hardware and hardware usage.
We need either
- to make sure that CI runs on data that forces batching 
- a way to manually make sure that the data is batched for testing even if there is more space available

SamTov:
In what way does it depend on hardware and what are the errors?

christophlohrmann:
In github CI everything passes.

On my laptop, one of the calculator tests fails reproducibly (I do not remember which one, but I can find out this evening)
It is a value-failure, so the calculation runs but produces wrong numerical values.

On my ICP machine, I can sometimes produce errors by using pytest -n 20 [pytest-xdist running tests in parallel with many runners]. Here I get an ``IndexError: list index out of range`` in a minibatch loop

SamTov:
In terms of the laptop fail please let me know tonight. There was a very recent change to a test that has now been removed and so if the repo was not updated this would fail but we can go through that.

As for the parallel nature of things #145 may the issue that we bring up here. I could imagine something funny happening if you had serious changes in memory availability but I don't have an exact idea of where. It is also just possible that some bug made it fail. But the fact that it fails when you run jobs in parallel is possible. When you run the parallel jobs, does it start 20 kernels and begin 20 projects or does it start one project, open a class and then run the tests on different threads? If there are shared variables in the parent but in a kernel something memory related gets changed then it could go wrong. But this could be corrected in the code by having dynamic, or as it is written in the issue, intelligent, memory management.

Ahh as I say that the mini batching may just not be working correctly. I would need to see an example and run it locally to confirm. But this atom-wise mini-batching may have been impacted by the restructure and now when you run 20 shells it invokes this mini-batching and so fails. But this would only happen on dynamic properties, which calculator failed?

christophlohrmann:
This time I got lucky and had one value error and two index errors:
```
FAILED integration_tests/calculators/test_radial_distribution_function.py::test_project - AssertionError: 

FAILED integration_tests/calculators/test_kirkwood_buff_integrals.py::test_project - tensorflow.python.framework.errors_impl.UnknownError: IndexError: list index out of range

FAILED integration_tests/calculators/test_radial_distribution_function.py::test_experiment - tensorflow.python.framework.errors_impl.UnknownError: IndexError: list index out of range
```

SamTov:
[...]

PR #447, opened 2021-12-15
Transformation refactor

### LINKED ISSUE #408: fix isort ciruclar import[database]
https://github.com/zincware/MDSuite/blob/17847d458e45ef84db37a509c018af1dd611ff45/mdsuite/database/__init__.py#L27-L31

christophlohrmann:
improve the transformations

todo:
- [x] move batching to transformations parent class
- [x] adapt all transformations to new parent
- [x] clean dependency structure of transformations that can trigger other transformations
- [ ] ~~implement generic batching along arbitrary (combination of) axis~~
- [x] test
- [x] doc

other improvements
- the names of the properties (e.g. Unwrapped_Positions) are now stored in only one place as a string and can therefore be swapped out easily. All other places use objects to talk about properties and their names. The strings are only required for the database.
- writing new transformations should be incredibly easy now
- testing is also streamlined since the actual transformation is separated from data handling (which has to be tested only once)
- remove all prints
- wrapping/unwrapping can handle time dependent box lengths if in the hdf database. In general, transformators will first use time-dependent data before falling back to constant values


christophlohrmann:
I will let the dust settle around #419  and then continue.
But some of the ideas are already demonstrated

christophlohrmann:
in the future, calculators should have the same 'property dependency' structure as the transformations, i.e. the calculator asks for the property and it gets fetched either from hdf, species properties, system properties, or triggers a transformation.

Before that is done, we should  first fix how (optional) simulation data (sample rate, temperature, box_l) are stored on the experiment and the species. For now, the transformations use workaround solutions like ``experiment.__getattribute__(property.name.lower())`` which could be eliminated if properties were collected in a single dict with PropertyInfo instance as keys.

dev-requirements.txt
```diff
@@ -1,3 +1,4 @@
+-r requirements.txt
```
PythonFZ: Not sure about that one

mdsuite/calculators/einstein_helfand_ionic_conductivity.py
```diff
@@ -182,10 +182,10 @@ def ensemble_operation(self, ensemble):
         MSD of the tensor_values.
         """
         msd = tf.math.squared_difference(
-            tf.gather(ensemble, self.args.tau_values, axis=0), ensemble[None, 0]
+            tf.gather(ensemble, self.args.tau_values, axis=1), ensemble[:, 0, :]
         )
-        msd = self.prefactor * tf.reduce_sum(msd, axis=1)
-        self.msd_array += np.array(msd)  # Update the averaged function
+        msd = self.prefactor * tf.reduce_sum(msd, axis=2)
+        self.msd_array += np.array(msd)[0, :]
```
PythonFZ, line 188: Is there any advantage of adding `arr[0, :]` over `arr[0]` ? I guess readability?

mdsuite/experiment/run.py
```diff
@@ -93,14 +94,14 @@ def transformation_wrapper(self, func: Union[Type[Transformations], Any]):
         func: a transformation to be attached to the experiment/s
         """
 
-        @functools.wraps(func)
+        @functools.wraps(func.run_transformation)
```
PythonFZ, line 97: Have we checked that the wraps works properly with autocompletion on the transformations?

mdsuite/experiment/run.py
```diff
@@ -93,14 +94,14 @@ def transformation_wrapper(self, func: Union[Type[Transformations], Any]):
...
         def wrapper(*args, **kwargs):
             if self.experiments is None:
                 self.experiments = [self.experiment]
             for experiment in self.experiments:
-                func_instance = func(*args, **kwargs)
+                func_instance = func()
                 # attach the transformation to the experiment
-                experiment.cls_transformation_run(func_instance)
+                experiment.cls_transformation_run(func_instance, *args, **kwargs)
             return None
```
PythonFZ: I guess the `return None` can be removed - my bad.

mdsuite/transformations/integrated_heat_current.py
```diff
@@ -24,125 +24,39 @@
...
-            batch_generator, args=batch_generator_args, output_signature=type_spec
+        super(IntegratedHeatCurrent, self).__init__(
+            input_properties=[
+                mdsuite_properties.unwrapped_positions,
+                mdsuite_properties.kinetic_energy,
+                mdsuite_properties.potential_energy,
+            ],
+            output_property=mdsuite_properties.integrated_heat_current,
+            batchable_axes=[1, 2],
+            scale_function={"linear": {"scale_factor": 2}},
```
PythonFZ, line 49: scale functions could also be a dataclass

mdsuite/transformations/integrated_heat_current.py
```diff
@@ -24,125 +24,39 @@
...
-        -------
-        Nothing.
-        """
-        self._compute_thermal_conductivity()  # run the transformation.
-        self.experiment.memory_requirements = self.database.get_memory_information()
+    def transform_batch(
+        self, batch: typing.Dict[str, typing.Dict[str, tf.Tensor]]
+    ) -> tf.Tensor:
+        currents = []
+        for _, properties in batch.items():
```
PythonFZ: `for properties in batch.values()`

mdsuite/transformations/ionic_current.py
```diff
@@ -24,186 +24,37 @@
...
-        -------
-        Nothing.
-        """
-        self._compute_ionic_current()  # run the transformation.
-        self.experiment.memory_requirements = self.database.get_memory_information()
+    def transform_batch(
+        self, batch: typing.Dict[str, typing.Dict[str, tf.Tensor]]
+    ) -> tf.Tensor:
+        currents = []
+        for _, properties in batch.items():
```
PythonFZ: same here

mdsuite/transformations/momentum_flux.py
```diff
@@ -24,135 +24,34 @@
...
-        -------
 
-        """
-        self._compute_momentum_flux()  # run the transformation.
-        self.experiment.memory_requirements = self.database.get_memory_information()
+    def transform_batch(
+        self, batch: typing.Dict[str, typing.Dict[str, tf.Tensor]]
+    ) -> tf.Tensor:
+        fluxes = []
+        for _, properties in batch.items():
```
PythonFZ: same here

mdsuite/transformations/thermal_flux.py
```diff
@@ -24,193 +24,71 @@
...
+                + np.multiply(stress[:, :, 1], vel[:, :, 1])
+                + np.multiply(stress[:, :, 5], vel[:, :, 2])
             )
             phi_z = (
-                np.multiply(data[stress_path][:, :, 4], data[velocity_path][:, :, 0])
-                + np.multiply(data[stress_path][:, :, 5], data[velocity_path][:, :, 1])
-                + np.multiply(data[stress_path][:, :, 2], data[velocity_path][:, :, 2])
+                np.multiply(stress[:, :, 4], vel[:, :, 0])
+                + np.multiply(stress[:, :, 5], vel[:, :, 1])
+                + np.multiply(stress[:, :, 2], vel[:, :, 2])
```
PythonFZ: I think `stress[:, :, 0] * vel[:, :, 0]` is easier to read than `np.multiply(stress[:, :, 0], vel[:, :, 0])` also feels like this could be mor vectorized but I would have to look up the formulas behind it. It is not really part of the PR but can be quickly changed.

mdsuite/transformations/thermal_flux.py
```diff
@@ -24,193 +24,71 @@
...
 
-            phi_sum = phi.sum(axis=0)
-            phi_sum_atoms = (
-                phi_sum / self.experiment.units["NkTV2p"]
-            )  # factor for units lammps nktv2p
+            phi_sum_atoms = phi.sum(axis=0)
+            # phi_sum_atoms = (
+            #         phi_sum_atoms / self.experiment.units["NkTV2p"]
+            # )  # factor for units lammps nktv2p
+            # TODO why is there a unit conversion in the transformation?
```
PythonFZ, line 87: This should be adressed! @SamTov do you know this?

mdsuite/utils/calculator_helper_methods.py
```diff
@@ -160,7 +163,7 @@ def func(x, m, a):
 #                 > self.experiment.number_of_configurations
 #                 - self.args.correlation_time
 #             ):
-#                 print("Trajectory not long enough to perform analysis.")
+#                 log.errorgit stat("Trajectory not long enough to perform analysis.")
```
PythonFZ: something went wrong here - not important because commented out

mdsuite/utils/meta_functions.py
```diff
@@ -316,7 +316,7 @@ def wrap(*args, **kw):
         ts = time()  # get the initial time
         result = f(*args, **kw)  # run the function.
         te = time()  # get the time after the function as run.
-        print(f"func:{f.__name__} took: {(te - ts)} sec")  # print the outcome.
+        log.info(f"func:{f.__name__} took: {(te - ts)} sec")
```
PythonFZ: I prefer `10 s` over `10 sec` in si units, what do you think?

PythonFZ (commented):
Difficult to review because so much changed. Looking through the code it looks much cleaner and less lines of code! I added some small comments to the PR and in parts to code that was only refactored.

mdsuite/experiment/run.py
christophlohrmann: no

mdsuite/transformations/integrated_heat_current.py
christophlohrmann: it could also be just a function (or lambda) then we do not need any intermediate agents to calculate memory consumption

mdsuite/utils/calculator_helper_methods.py
christophlohrmann: good catch. looks like my cursor was in the wrong window :)

dev-requirements.txt
christophlohrmann: saves one pip install for devs

christophlohrmann:
done so far. tomorrow I will address the review points already mentioned and this should be good to go

christophlohrmann:
some of the transformations remain untested (see CI/unit_tests/transformations/test_transformations.py) these are the ones I do not understand and where I would invite the people who wrote them initially to come in and make sure nothing is broken.

dev-requirements.txt
PythonFZ: yeah, but as dev you will do `pip install -e .` and `pip install -r dev-req.txt` so the first one will always install the `requierments.txt` I think it would make more sense to have the setup.py use a proper dev install that does both

mdsuite/transformations/integrated_heat_current.py
PythonFZ: A lambda function would be nice and quite flexible. @SamTov iirc you added that, do you think something like `scale_function=lambda x: 2*x` would also work?

dev-requirements.txt
christophlohrmann: you are right
my change was propmted by the frustration of failing tests bcs of a new dependency in req.txt that I did not get to work by rerunning pipi install dev-req. Maybe a bit of an overreaction on my part

mdsuite/transformations/integrated_heat_current.py
SamTov: Yeah I think I prefer the data class for a reduction in code re-implementation. For all calculators, transformations, and anything loading data we have some scale function. Of course, it isn't a lot of additional code but if we have a data-class or these dicts then you don't have to re-write the function each time. 

If we think that having a lambda on each one is better than having them all sitting in one place then we can do that, but it is in any case for another PR.

mdsuite/transformations/thermal_flux.py
SamTov: It is more a scaling factor from memory not a direct unit conversion. It should retain the units. But this is from @Fratorhe so he should confirm that.

mdsuite/database/simulation_data_class.py
```diff
@@ -43,8 +43,8 @@ class SimulationProperties:
     thermal_flux = ("Thermal_Flux", (None, 3))
     stress_viscosity = ("Stress_visc", (None, 3))
     momentum_flux = ("Momentum_Flux", (None, 3))
-    ionic_current = ("Ionic_Current", (None, 3))
-    translational_dipole_moment = ("Translational_Dipole_Moment", (None, 3))
+    ionic_current = ("Ionic_Current", (None, None, 3))
```
SamTov: Why is this different from the momentum flux for example? It is still a system-wide property.

mdsuite/memory_management/memory_manager.py
```diff
@@ -212,12 +212,12 @@ def get_batch_size(self, system: bool = False) -> tuple:
...
                 1,
-                n_columns - self.offset,
+                n_configs - self.offset,
             )
         )
         batch_size = self._get_optimal_batch_size(maximum_loaded_configurations)
-        number_of_batches = int((n_columns - self.offset) / batch_size)
-        remainder = int(n_columns % batch_size)
+        number_of_batches = int((n_configs - self.offset) / batch_size)
+        remainder = int(n_configs % batch_size)  # smells fishy: no offset
```
SamTov: I think it definitely needs an offset.

mdsuite/transformations/ionic_current.py
```diff
@@ -24,186 +24,38 @@
 Summary
```
SamTov, line 24: Just to confirm as I did not see a anything explicity, this can handle atom-wise charges dumped during a simulation?

mdsuite/transformations/velocity_from_positions.py
```diff
@@ -0,0 +1,59 @@
...
+
+from mdsuite.database.simulation_data_class import mdsuite_properties
+from mdsuite.transformations.transformations import SingleSpeciesTrafo
+
+
+class VelocityFromPositions(SingleSpeciesTrafo):
+    """
+    Calculate the velocity based on the particle positions via simple forward derivative,
+    i.e. v(t) = (x(t+dt)-x(t))/dt.
+    The last velocity of the trajectory cannot be computed and is copied
```
SamTov, line 26: How much of an effect will this have on the calculations? One could imagine, for example, an auto-correlation function having quite the peak at a false t value given essentially a static flux.

SamTov (approved):
Looks great, I made a few comments but just general questions. One important point though is about the shapes of some of the data. Have we changed anything about stored data or how it is accessed shape-wise?

mdsuite/database/simulation_data_class.py
christophlohrmann: In the hdf database, everything has dimension 3, where the particle axis has length 1 for system properties. That simplifies all data storing/loading and also makes it such that tensorflow does not have to complain about dimension mismatch when combining system properties and particle properties.
The dict that you marked here should go away soon, the transformations do not use it anymore.
I changed the properties (ionic current) that are generated by the new transformations but used by the 'old' calculators.

mdsuite/transformations/ionic_current.py
christophlohrmann: the nice bit is that as a transformation user, you just request a certain property for the calculation.
the transformation parent will give you, depending on availability
1. time dependent data
2. data constant for the species
3. data constant for the whole system
So if there are charges dumped from simulation, you will get the (n_part,n_config,1) array of charges.
If not, you get a (1,1,1) array of the charge of that species. Tensorflow broadcasting makes it such that the calculation does not care which is actually there.

mdsuite/transformations/velocity_from_positions.py
christophlohrmann: I think if you do not have 'instant' velocity data and need to compute it from the positions, you need to have a fine enough resolution to capture all acceleration/deceleration of particles for this transformation to make sense. And if that resolution is good enough, autocorrelations should not suffer from a single slightly wrong value

mdsuite/memory_management/memory_manager.py
christophlohrmann: #383 :)

mdsuite/transformations/thermal_flux.py
SamTov: I think if the data is all tf.Tensor then using np might actually slow it down unless tf automatically does the conversion.

mdsuite/database/simulation_data_class.py
SamTov: Perfect thanks.

mdsuite/memory_management/memory_manager.py
SamTov: True...

mdsuite/transformations/ionic_current.py
SamTov: Excellent.

mdsuite/transformations/velocity_from_positions.py
SamTov: That sounds reasonable but perhaps this could be kept in mind. I just see for an autocorrelation function that for the last two time steps:

v(t) * v(t + 1) == 1,

which is probably just off of what it should be. However, as you mentioned for correct resolution of steps this should not cause any serious issue.

mdsuite/transformations/thermal_flux.py
christophlohrmann: I changed it to standard tf multiplication

mdsuite/database/simulation_data_class.py
christophlohrmann: 17e3ad6 removes the None - business

PythonFZ (approved):

mdsuite/experiment/run.py
christophlohrmann: checked it manually: autocomplete works

PR #448, opened 2021-12-16
add scaling functions unit tests

SamTov:
Added unit tests for the scaling functions module.


SamTov:
I think the flake8 stuff is handled in the transformations API change. At least most of it anyway.

CI/integration_tests/transformations/test_unwrap_via_indices.py
```diff
@@ -0,0 +1,64 @@
...
+        return (temporary_path / NaCl.file_raw).as_posix()
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        """
+        Move out of tmp directory and delete it at end of test.
+        Returns
+        -------
+
+        """
```
PythonFZ: Are you missing a commit here, which adds the test?

CI/unit_tests/utils/test_scaling_functions.py
```diff
@@ -0,0 +1,131 @@
...
+)
+
+
+class TestScalingFunctions(unittest.TestCase):
+    """
+    A test class for the scaling functions module.
+    """
+
+    def test_linear_scaling_function(self):
+        """
```
PythonFZ, line 41: I think for all the scaling tests it would be better to test them against a function that actually scales like this. 
Currently you are only testing that the numpy code you wrote works. This ensures that no one accidentally changes the code but does not ensure that the scaling actually is the scaling we are interested in.

PythonFZ (changes requested):
The empty test should be added / removed.

CI/integration_tests/transformations/test_unwrap_via_indices.py
SamTov: I guess so... but I thought I took this off main. Was this test added in the transformation API? I will double check I have merged correctly with main and that it is updated though

CI/unit_tests/utils/test_scaling_functions.py
SamTov: I compute the value by hand as it is a simple algebraic expression. The only time I did not do this was for the log one. I didn't use the output of the function as the test comparison I wrote the function out, computed it on the values and then tested that those numbers came out of the tests. We could write it such that it is tested on random numbers and then use a written polynomial for each but I think this suffices.

CI/integration_tests/transformations/test_unwrap_via_indices.py
SamTov: This branch is up to date with main.

PythonFZ (approved):

PR #450, opened 2021-12-18
SamTov update visualiser

SamTov:
Following additions:

* Add ZnVis visualizer
* Add example for visualizer
* Add integration test for visualizer
* Add MDSuiteProcess class for multi-process tests
* Add tests for MDSuiteProcess class
* Add test for assertDeepAlmostEqual functionality

## Important
This PR can be reviewed but not merged until ZnVis is made public. For this reason the visualizer test will currently fail. I would expect ZnVis to be public by next week.


PythonFZ:
Just read it on the phone, but ZnVis is missing in the requirements.txt file. Will do a full review later.

SamTov:
> Just read it on the phone, but ZnVis is missing in the requirements.txt file. Will do a full review later.

Yes, that was the idea behind the PR comment under the headline, important. I will update the package stuff when ZnVis is public.

PythonFZ:
> > Just read it on the phone, but ZnVis is missing in the requirements.txt file. Will do a full review later.
> 
> Yes, that was the idea behind the PR comment under the headline, important. I will update the package stuff when ZnVis is public.

I missed that on mobile. I'd suggest converting this PR into a draft then and add add a review when ready.

SamTov:
> > > Just read it on the phone, but ZnVis is missing in the requirements.txt file. Will do a full review later.
> > 
> > 
> > Yes, that was the idea behind the PR comment under the headline, important. I will update the package stuff when ZnVis is public.
> 
> I missed that on mobile. I'd suggest converting this PR into a draft then and add add a review when ready.

The idea is that the code passes all tests locally and is just waiting for the other package to be made public. For this reason I put it up for review as it cannot merge until that package is public however the code here will not change. It is, in this way, completely ready for review.

CI/integration_tests/visualizer/znvis_visualizer.py
```diff
@@ -0,0 +1,94 @@
...
+        TODO: We can set up a means of havng the main thread check if the test is
+            finished on the deployed process and then kill it. That will ensure that the
+            test isn't prematurely closed and passes when it should have failed. For now
+            120 seconds is more than enough time for this test but we can think of some
+            improvements later.
+        """
+        process = MDSuiteProcess(target=self._run_app)
+        process.start()
+        time.sleep(90)  # give it 90 seconds to run.
+        process.terminate()
```
PythonFZ, line 63: Why 90 s?

CI/unit_tests/utils/test_testing.py
```diff
@@ -0,0 +1,109 @@
...
+        Returns
+        -------
+
+        """
+        process = MDSuiteProcess(target=self._exception_throw)
+        process.start()
+        time.sleep(5)
+        process.terminate()
+        self.assertEqual(True, process.exception is not None)
+
```
PythonFZ: `assert True` should be sufficient instead of `assert True == True`.

CI/unit_tests/utils/test_testing.py
```diff
@@ -0,0 +1,109 @@
...
+        Returns
+        -------
+
+        """
+        process = MDSuiteProcess(target=self._no_exception)
+        process.start()
+        time.sleep(5)
+        process.terminate()
+        self.assertEqual(True, process.exception is None)
+
```
PythonFZ: see above

mdsuite/calculators/spatial_distribution_function.py
```diff
@@ -249,10 +249,9 @@ def run_calculator(self):
             subjects=["System"],
         )
 
-        if self.plot:
-            coordinates = tf.reshape(self._get_unit_sphere(), [self.args.n_bins ** 2, 3])
-            colour_map = tf.reshape(sdf_values, [-1])
-            self._run_visualization(coordinates, colour_map)
+        coordinates = tf.reshape(self._get_unit_sphere(), [self.args.n_bins ** 2, 3])
+        colour_map = tf.reshape(sdf_values, [-1])
+        self._run_visualization(coordinates, colour_map)
```
PythonFZ, line 254: Why remove the `plot=True` argument? The SDF does store data in the DB https://github.com/zincware/MDSuite/blob/244a4ca81d27599e81487ea51a2d81473683e8f0/mdsuite/calculators/spatial_distribution_function.py#L246-L247 so it should have a `plot=False` option?

mdsuite/utils/testing.py
```diff
@@ -58,6 +61,40 @@ def assertDeepAlmostEqual(expected, actual, *args, **kwargs):
...
+class MDSuiteProcess(multiprocessing.Process):
+    """
+    Process class for use in ZnVis testing.
+    """
+
+    def __init__(self, *args, **kwargs):
+        """
+        Multiprocessing class constructor.
+        """
+        multiprocessing.Process.__init__(self, *args, **kwargs)
```
PythonFZ: This should be a super call.

mdsuite/utils/testing.py
```diff
@@ -58,6 +61,40 @@ def assertDeepAlmostEqual(expected, actual, *args, **kwargs):
         assert expected == actual
 
 
+class MDSuiteProcess(multiprocessing.Process):
+    """
+    Process class for use in ZnVis testing.
+    """
```
PythonFZ, line 67: Can you explain this class and the associated tests to me? I don't really understand what it does tbh.

mdsuite/visualizer/znvis_visualizer.py
```diff
@@ -0,0 +1,142 @@
...
+        colour : np.ndarray
+                RBG array of colours.
+        radius : float
+                Radius of the particles. This is a reduced mass.
+        """
+        # Load the species data from pubchempy data file.
+        data_name = "mdsuite.data"
+        json_name = "PubChemElements_all.json"
+        with importlib.resources.open_text(data_name, json_name) as json_file:
+            pse = json.loads(json_file.read())
```
PythonFZ: I'd prefer using the following, because it will support refactoring:

```py
import mdsuite.data

pse = json.loads(importlib.resources.read_text(mdsuite.data, "PubChemElements_all.json"))
```

PythonFZ (changes requested):
Looks good. A few comments and I can only approve the PR when the requirements.txt is added.

CI/integration_tests/visualizer/znvis_visualizer.py
SamTov: I do explain why there is a time here in the TODO. It is a different time in the todo but the concept is the same. 

The point here is that after enough time for the test to run the process must be killed. Right now I don't include communication from the process back to the main test and so I simply wait for 90 seconds, enough time for the test to run, and then kill it. In future, we can work on a way to communicate between the threads but as long as the process test is not long this is fine.

mdsuite/calculators/spatial_distribution_function.py
SamTov: When plot is run it calls a method that the SDF does not have or want to have. We need to remove that or we can leave it at plot=False. As the SDF is less of a calculator and more visualisation tool I don't think it needs a plot method.

mdsuite/visualizer/znvis_visualizer.py
SamTov: yep that is a good point.

mdsuite/utils/testing.py
SamTov: Absolutely. If there is a better way to handle this also please let me know. 

When we want to run a test calling Open3D we need to open it. Open3D will run a separate process as it allows for callbacks but this means that it cannot be forcefully closed during it's run, it must be closed by a user. The idea behind using multiprocessing for a test is that we can open the code in a new process, let it run and then use `process.terminate()` to forcefully end it. The reason the Process is sub-classed is because we want to use it for a test and therefore need a way of collecting the exceptions or errors that were raised on the deployed process. This class collects these methods such that they can be looked at in the test. Without this functionality, if an exception is raised on the deployed process all of the tests will still pass.

CI/unit_tests/utils/test_testing.py
SamTov: resolved in d8c6389

CI/unit_tests/utils/test_testing.py
SamTov: resolved in d8c6389

mdsuite/visualizer/znvis_visualizer.py
SamTov: resolved in 71ff782

CI/unit_tests/utils/test_testing.py
```diff
@@ -0,0 +1,111 @@
...
+        -----
+        Taken from the module __main__
+        """
+        dict_2a = {"a": {"b": np.array([1, 2, 3, 4])}}
+        dict_2b = {"a": {"b": [1, 2, 3, 4]}}
+        dict_3a = {"a": {"c": np.array([1.10, 2.10, 3.11, 4.0])}}
+        dict_3b = {"a": {"c": np.array([1.11, 2.09, 3.10, 4.0])}}
+
+        print(assertDeepAlmostEqual(dict_3a, dict_3b, decimal=1))
+        print(assertDeepAlmostEqual(dict_2a, dict_2b, decimal=1))
```
PythonFZ: No need for the print statement, I guess.

mdsuite/utils/testing.py
SamTov: resolved in 71ff782

SamTov: resolved in eb448fc

CI/unit_tests/utils/test_testing.py
SamTov: good point, I just copied and pasted from the module and did not read that test

SamTov: resolved in eb448fc

SamTov:
I think this is resolved now.

PythonFZ (approved):
Maybe remove the SDF example script for now and extend the docs. But looks good!

SamTov (commented):
Checking for spurious file addition. None found after conflict resolution.

PR #451, opened 2021-12-18
Colab

PythonFZ:
- Add `open in colab` badge to example notebook
- Add pip install

# [Binder](https://mybinder.org/) might be the better option

Checkout https://github.com/zincware/MDSuite/blob/colab/examples/notebooks/NaCl_walkthrough.ipynb for the added changes.

# TODO
- [ ] before merge change `https://colab.research.google.com/github/zincware/MDSuite/blob/colab/examples/notebooks/NaCl_walkthrough.ipynb` into `https://colab.research.google.com/github/zincware/MDSuite/blob/main/examples/notebooks/NaCl_walkthrough.ipynb`
- [ ] fix the issues in the colab notebook
- [ ] either set `config.jupyter  = True` or find a way to dectec colab notebooks, because the jupyter detection does not seem to work.


PythonFZ:
Use binder instead of colab

PR #452, opened 2021-12-21
Small fixes for SimulationDatabase

### LINKED ISSUE #400: simulation_database should use a context manager
The `simulation_database.py` often uses open and close statements.
We should try to avoid this because as far as I know HDF5 DBs are likely to corrupt when not closed properly.

We should replace every `hf.File()` with a context manager for safety.

### LINKED ISSUE #449: pathlib support in simulation_database
I think that `name` is a misleading argument, because as far as I understand it is the path to the database file.
I would also like to make it pathlib compatible, so we don't have to use `path.as_posix()`, which I assume hf.File unfortunately does not support natively, but we should check that.

https://github.com/zincware/MDSuite/blob/c7bb82667012ed9d274551314f659596e7f730e5/mdsuite/database/simulation_database.py#L221-L236

PythonFZ:
Fix #449 
Fix #400

# Summary

- remove `Database.architecture`
- rename `Database.name` to `Database.path`

### DIFF
--- a/mdsuite/database/simulation_database.py
+++ b/mdsuite/database/simulation_database.py
@@ -28,8 +28,8 @@ import dataclasses
 import logging
 import pathlib
 import time
-from typing import List
 import typing
+from typing import List
 
 import h5py as hf
 import numpy as np
--- a/mdsuite/database/simulation_database.py
+++ b/mdsuite/database/simulation_database.py
@@ -228,7 +228,14 @@ class Database:
         path : str|Path
                 The name of the database_path in question.
         """
-        self.path = pathlib.Path(path).as_posix()  # name of the database_path
+        if isinstance(path, pathlib.Path):
+            self.path = path.as_posix()
+        elif isinstance(path, str):
+            self.path = path  # name of the database_path
+        else:
+            # TODO fix this!
+            log.debug(f"Expected str|Path but found {type(path)}")
+            self.path = path
 
     @staticmethod
     def _update_indices(
--- a/mdsuite/database/simulation_database.py
+++ b/mdsuite/database/simulation_database.py
@@ -225,27 +225,10 @@ class Database:
 
         Parameters
         ----------
-        path : str
+        path : str|Path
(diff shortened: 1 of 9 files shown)


mdsuite/database/simulation_database.py
```diff
@@ -210,47 +209,27 @@ class Database:
...
-        Closes the database_path object
-        """
-
-        database.close()
+        if isinstance(path, pathlib.Path):
+            self.path = path.as_posix()
+        elif isinstance(path, str):
+            self.path = path  # name of the database_path
+        else:
+            # TODO fix this!
```
SamTov, line 230: what fix are you looking for?

mdsuite/time_series/base.py
```diff
@@ -26,7 +26,6 @@
 """
```
SamTov, line 26: are these used anywhere? Should we bring this under the calculators at some stage?

SamTov (approved):
Everything look nice. I added a few tests for the simulation_database to build on the coverage. There are two generic question in the review but not relevant for the merge.

mdsuite/database/simulation_database.py
PythonFZ: Related to #453 this should only take str or pathlib and not <class byte>

mdsuite/time_series/base.py
PythonFZ: They are probably not used at the moment so we should look into either removing or moving them

mdsuite/database/simulation_database.py
SamTov: ahh okay.

mdsuite/time_series/base.py
SamTov: They are nice to have as we can compute a moving average on properties which is very useful for pressure, energy, temperature, and so on. I think we should consider how to move them into the code under run with something like a `MovingAverage` calculator or something.

PR #457, opened 2022-01-05
Sam tov correlation coefficients

SamTov:
Fixed the distinct diffusion coefficient calculators and brought them under the new calculator framework.

* Fix GK distinct diffusion coefficients
* Fix Einstein distinct diffusion coefficients
* Update integration tests and add them back to the suite.


mdsuite/calculators/trajectory_calculator.py
```diff
@@ -367,8 +369,10 @@ def get_ensemble_dataset(self, batch: dict, subject: str):
         )
 
         type_spec = {}
-
-        loop_list = [subject]
+        if type(subject) is str:
```
PythonFZ: should be `if isinstance(subject, str)` see https://switowski.com/blog/type-vs-isinstance

CI/integration_tests/calculators/test_einstein_distinct_diffusion_coefficients.py
```diff
@@ -54,52 +56,52 @@ def true_values() -> dict:
...
-        r"C:\Users\fabia\Nextcloud\DATA\JupyterProjects\MDSuite\CI\integration_tests"
-        r"\calculators\data\einstein_distinct_diffusion_coefficients.json"
-    )
-
-    data.write_text(json.dumps(data_dict))
-
-    np.testing.assert_array_almost_equal(data_dict["x"], true_values["x"])
-    np.testing.assert_array_almost_equal(
-        data_dict["uncertainty"], true_values["uncertainty"]
+        plot=False, data_range=300, correlation_time=100
```
PythonFZ, line 68: we should add data to compare the results to to the datahub

mdsuite/calculators/einstein_distinct_diffusion_coefficients.py
```diff
@@ -170,61 +202,24 @@ def _compute_msd(self, data: dict, data_path: list, combination: tuple):
...
-            for batch in data_set:
-                self._compute_msd(batch, data_path, combination)
-            self._apply_averaging_factor()
-            self._post_operation_processes(combination)
+        msd_a = self.msd_operation(data[data_path[0]], square=False)
+        msd_b = self.msd_operation(data[data_path[0]], square=False)
+
+        for i in range(len(data[data_path[0]])):
+            for j in range(i + 1, len(data[data_path[1]])):
+                if combination[0] == combination[1] and i == j:
```
PythonFZ, line 220: missing docstring for combination

mdsuite/calculators/einstein_distinct_diffusion_coefficients.py
```diff
@@ -170,61 +202,24 @@ def _compute_msd(self, data: dict, data_path: list, combination: tuple):
...
-            data_set = tf.data.Dataset.from_generator(
-                batch_generator, args=batch_generator_args, output_signature=type_spec
-            )
-            data_set = data_set.prefetch(tf.data.experimental.AUTOTUNE)
-            for batch in data_set:
-                self._compute_msd(batch, data_path, combination)
-            self._apply_averaging_factor()
-            self._post_operation_processes(combination)
+        msd_a = self.msd_operation(data[data_path[0]], square=False)
+        msd_b = self.msd_operation(data[data_path[0]], square=False)
```
PythonFZ, line 216: Shapes would be helpful here

mdsuite/calculators/einstein_distinct_diffusion_coefficients.py
```diff
@@ -170,61 +202,24 @@ def _compute_msd(self, data: dict, data_path: list, combination: tuple):
...
+
+        for i in range(len(data[data_path[0]])):
+            for j in range(i + 1, len(data[data_path[1]])):
+                if combination[0] == combination[1] and i == j:
+                    continue
+                else:
+                    self.msd_array += self.prefactor * np.array(
+                        tf.reduce_sum(msd_a[i] * msd_b[j], axis=1)
+                    )
```
PythonFZ: isn't this whole thing something like `tf.einsum("ixy, jxy -> x", msd_a, msd_b) - <the diagonal entries>` - how important is this part for the performance?

PythonFZ (changes requested):
some small changes

CI/integration_tests/calculators/test_einstein_distinct_diffusion_coefficients.py
SamTov: I definitely agree. However, as with all of the tests here, we should decide on the data we use for the comparison to make sure it is a descriptive test. For now, these tests are more to identify faults in the pipeline.

mdsuite/calculators/trajectory_calculator.py
SamTov: resolved in 7f802b9

mdsuite/calculators/einstein_distinct_diffusion_coefficients.py
SamTov: Resolved in c4ac13d

mdsuite/calculators/einstein_distinct_diffusion_coefficients.py
SamTov: resolved in 92c5681

mdsuite/calculators/einstein_distinct_diffusion_coefficients.py
SamTov: It is certainly important for performance and definitely can be reduced to a summation over everything and then subtraction of diagonals but in the above einsum I am a little confused with where your tensor contractions are taking place. Could you provide an array example of what this looks like because by tensor contraction laws this should be a sum over x and y and the i and j should remain which is not quite what we want..

SamTov: In any case, there are two or three approaches I want to try for a speedup in the distinct coefficients. It may be better if I open an issue for improving correlation coefficient speed and when we have some nice test data play around with improving performance.

PythonFZ: Issue sounds good - we can collect potential methods there and  compare to existing data. Can be resolved when the issue is made.

PythonFZ (approved):
Looks all good!

PR #469, opened 2022-01-18
fix scale coordinates trafo

### LINKED ISSUE #459: position property naming
currenty we have 
```
Positions
Scaled_Positions
Unwrapped_Positions
Scaled_Unwrapped_Positions
```

I think there is quite the potential for confusion when calculators and transformations work with positions, thinking they are already scaled (I am pretty sure all occurences of Unwrapped_Positions actually would like the Scaled_Unwrapped_Positions). I would therefore propose a renaming to:

```
Normalized_Positions
Positions
Normalized_Unwrapped_Positions
Unwrapped_Positions
```

We could even go as far as making Unwrapped the standard and write Wrapped_** for the others.

christophlohrmann:
Fix #459 
 does not rename, instead fix the transformation


CI/unit_tests/transformations/test_transformations.py
```diff
@@ -96,16 +96,17 @@ def test_momentum_flux():
...
     trafo = scale_coordinates.ScaleCoordinates()
-    assert trafo.output_property == mdsuite_properties.scaled_positions
+    assert trafo.output_property == mdsuite_properties.positions
 
     n_part = 5
     n_step = 7
 
     pos = tf.convert_to_tensor(np.random.random((n_part, n_step, 3)), dtype=dtype)
+    pos /= np.max(pos)
     box_l = tf.convert_to_tensor([1.1, 2.2, 3.3], dtype=dtype)[None, None, :]
```
PythonFZ, line 106: it's a test so not imporant but `tf.random.normal(())` and `pos /= tf.reduce_max(pos)` would keep everything in the TensorFlow Framework

mdsuite/database/mdsuite_properties.py
```diff
@@ -64,8 +64,8 @@ class MDSuiteProperties:
     kinaci_heat_current = PropertyInfo("Kinaci_Heat_Current", 3)
     charge = PropertyInfo("Charge", 1)
     energy = PropertyInfo("Energy", 1)
-    kinetic_energy = PropertyInfo("KE", 1)  # TODO rename, also in calculators
-    potential_energy = PropertyInfo("PE", 1)  # TODO rename, also in calculators
+    kinetic_energy = PropertyInfo("Kinetic_Energy", 1)
+    potential_energy = PropertyInfo("Potential_Energy", 1)
```
PythonFZ, line 81: this has been changed in the calculators prior to this to use the dataclass?

mdsuite/transformations/transformation_dict.py
```diff
@@ -84,13 +84,15 @@
     mdp.ionic_current: ionic_current.IonicCurrent,
     mdp.kinaci_heat_current: kinaci_integrated_heat_current.KinaciIntegratedHeatCurrent,
     mdp.momentum_flux: momentum_flux.MomentumFlux,
-    mdp.scaled_positions: scale_coordinates.ScaleCoordinates,
+    mdp.positions: [
+        scale_coordinates.ScaleCoordinates,
+        wrap_coordinates.CoordinateWrapper,
```
PythonFZ: I'm not sure about this whole file but I guess it is necessary.

One thing that we might want to change would be 
`from mdsuite.transformations import unwrap_coordinates; unwrap_coordinates.CoordinateUnwrapper` 
to add all of them to the `mdsuite.transformations.__init__` to the `__all__` and then simply have `from mdsuite.transformations import CoordinateUnwrapper` ?

PythonFZ (changes requested):
Can we add some Docstrings to ` MDSuiteProperties` that explain `scaled_positions` and `positions` (ideally all of them if not 100 % clear, but not part of the PR)? Probably via `Attributes` in the docstring?

mdsuite/database/mdsuite_properties.py
christophlohrmann: the calculators already use the dataclass member. The strings only show up here

PythonFZ: perfect! Very nice change!

mdsuite/transformations/transformation_dict.py
christophlohrmann: sadly I have not found a way to circumvent this, I know it is bad design.
I am not a fan of the __all__ mechanism (bcs it uses strings to point to classes which I hate).
Since the transformations are very small now anyway, we could just put them all in transformations.py to make importing easier

CI/unit_tests/transformations/test_transformations.py
christophlohrmann: I'm still learning tf and for the tests I went with the stuff I know works

mdsuite/transformations/transformation_dict.py
PythonFZ: Have you tried `__all__ = [CoordinateUnwrapper.__name__, ...]` ?

christophlohrmann: nope. but that sounds much better already

SamTov (approved):
I don't have any further comments on the code. It looks great, thank you for updating it.

PythonFZ (approved):

PR #470, opened 2022-01-19
Bypass generator minibatching in RDF/ADF calculator.

SamTov:
Forcefully bypass the generator mini batching for RDF and ADF as it is built into the calculator itself. This resolves issues that can arise when mini batching is triggered on these calculators. Not sure yet whether this will solve the random test failures.

### DIFF
--- a/mdsuite/calculators/angular_distribution_function.py
+++ b/mdsuite/calculators/angular_distribution_function.py
@@ -215,7 +215,9 @@ class AngularDistributionFunction(TrajectoryCalculator, ABC):
         self.gpu = gpu
         self.plot = plot
         self._batch_size = batch_size  # memory management for all batches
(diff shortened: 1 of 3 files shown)


PythonFZ (approved):
I guess for now - if it works this hopefully resolves some issues. We should add some benchmakrs, e.g. https://pytest-benchmark.readthedocs.io/en/latest/ at some point in the future for changes that affect performance.

SamTov:
> I guess for now - if it works this hopefully resolves some issues. We should add some benchmakrs, e.g. https://pytest-benchmark.readthedocs.io/en/latest/ at some point in the future for changes that affect performance.

This one should not impact the performance of the calculators but that is a good idea for the CI in general.

PR #467, opened 2022-01-18
Remove SDF Code

PythonFZ:
Fix Issues with open3D e.g. in binder

- ZnVIS and open3D were either commented out or removed from the MDSuite code.
- SDF Calculator was removed from the calculators imported by default
- ZnVIS and open3d were removed from the `requirements.txt`

PR #472, opened 2022-01-20
Sam tov zn vis

SamTov:
Make visualizer an optional add-on that requires installing the ZnVis package separately. At the same time, remove the SDF code from the package using comments to avoid loss of features.

Has been tested on a fresh environment without ZnVis package and it raises the required logging statement and prevent the visualizer from being called.


.github/workflows/pytest.yml
```diff
@@ -35,6 +35,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -r dev-requirements.txt
+          pip install znvis
```
PythonFZ: Why not put it into dev-requirements.txt?

PythonFZ:
I think we should merge this with  #467 I don't think we need to branches for that.

SamTov:
> I think we should merge this with #467 I don't think we need to branches for that.

Yep but because a lot of vis stuff was removed in #467 and it is probably easier not to go back and re-add it we can probably just close that and use this PR.

PythonFZ:
> > I think we should merge this with #467 I don't think we need to branches for that.
> 
> Yep but because a lot of vis stuff was removed in #467 and it is probably easier not to go back and re-add it we can probably just close that and use this PR.

There are some parts in the other PR that weren't used in MDSuite anymore so I would like to merge them.

SamTov:
> > > I think we should merge this with #467 I don't think we need to branches for that.
> > 
> > 
> > Yep but because a lot of vis stuff was removed in #467 and it is probably easier not to go back and re-add it we can probably just close that and use this PR.
> 
> There are some parts in the other PR that weren't used in MDSuite anymore so I would like to merge them.

Yep that's no problem for me just have to be careful that none of the stuff is removed. I'd like to keep basically everything related to the SDF just have it commented out.

PythonFZ (approved):
Works in Binder - looks good!

.github/workflows/pytest.yml
SamTov: Thought about it, and don't mind the idea, I left it out as I figured it is not explicitly a dev requirement. But if you think it also can go there then I can move it.

PythonFZ: I would put it there, because it is required for the tests to run. If you run them locally they will fail otherwise.

SamTov: solved in 57b988f

PR #473, opened 2022-01-21
GPU batching

### LINKED ISSUE #336: make `GPU` a global setting
Currently the GPU args in the calculators does not disable the GPU but only affects the batching.
We should make it a `mds.config.GPU = True`global option and look at that.

Also we should go through all the calculators and check the arguments

PythonFZ:
- fix #336 (partially, but I think it is resolved then)

Should we add something like

```py
import os
os.environ["CUDA_VISIBLE_DEVICES"] = "-1"
```

to the docs to let people know how to disable GPU?
Maybe we could also have something like `python mdsuite --disable-gpu` but I have not looked into that.


SamTov (commented):
I think this is great, however, before we merge it the scaling functions need to be addressed. In my recent attempts at using MDSuite on larger dataset (~100 GB), the scaling functions are not adequate for the smaller memory systems. When we have something like a 90 GB memory system it can absorb some of the failures of the scaling functions but for the GPUs there is no room for this kind of error. Therefore, before making the GPU default, which will almost certainly result in memory failures immediately for those analysing big systems, the scaling functions must be fit correctly.

PythonFZ:
>I think this is great, however, before we merge it the scaling functions need to be addressed. In my recent attempts at using MDSuite on larger dataset (~100 GB), the scaling functions are not adequate for the smaller memory systems. When we have something like a 90 GB memory system it can absorb some of the failures of the scaling functions but for the GPUs there is no room for this kind of error. Therefore, before making the GPU default, which will almost certainly result in memory failures immediately for those analysing big systems, the scaling functions must be fit correctly.

I do think, that we can and should merge this anyway. It does not affect the issue you mention, because GPU is default anyways.
When a GPU is available TensorFlow will use it, so this at least helps in some cases.

SamTov:
> > I think this is great, however, before we merge it the scaling functions need to be addressed. In my recent attempts at using MDSuite on larger dataset (~100 GB), the scaling functions are not adequate for the smaller memory systems. When we have something like a 90 GB memory system it can absorb some of the failures of the scaling functions but for the GPUs there is no room for this kind of error. Therefore, before making the GPU default, which will almost certainly result in memory failures immediately for those analysing big systems, the scaling functions must be fit correctly.
> 
> I do think, that we can and should merge this anyway. It does not affect the issue you mention, because GPU is default anyways. When a GPU is available TensorFlow will use it, so this at least helps in some cases.

Perhaps I did not state it very well but it is completely involved in the issue I mentioned as far as I can see. Say you have GPU as default and try to run analysis on the 40 GB molten salt data, RDFs and GK diffusion for example both fail due to memory. This means that our default parameter is not memory safe.

PythonFZ:
> > > I think this is great, however, before we merge it the scaling functions need to be addressed. In my recent attempts at using MDSuite on larger dataset (~100 GB), the scaling functions are not adequate for the smaller memory systems. When we have something like a 90 GB memory system it can absorb some of the failures of the scaling functions but for the GPUs there is no room for this kind of error. Therefore, before making the GPU default, which will almost certainly result in memory failures immediately for those analysing big systems, the scaling functions must be fit correctly.
> > 
> > 
> > I do think, that we can and should merge this anyway. It does not affect the issue you mention, because GPU is default anyways. When a GPU is available TensorFlow will use it, so this at least helps in some cases.
> 
> Perhaps I did not state it very well but it is completely involved in the issue I mentioned as far as I can see. Say you have GPU as default and try to run analysis on the 40 GB molten salt data, RDFs and GK diffusion for example both fail due to memory. This means that our default parameter is not memory safe.

But doesn't it fail with the current version as well?

SamTov:
> > > > I think this is great, however, before we merge it the scaling functions need to be addressed. In my recent attempts at using MDSuite on larger dataset (~100 GB), the scaling functions are not adequate for the smaller memory systems. When we have something like a 90 GB memory system it can absorb some of the failures of the scaling functions but for the GPUs there is no room for this kind of error. Therefore, before making the GPU default, which will almost certainly result in memory failures immediately for those analysing big systems, the scaling functions must be fit correctly.
> > > 
> > > 
> > > I do think, that we can and should merge this anyway. It does not affect the issue you mention, because GPU is default anyways. When a GPU is available TensorFlow will use it, so this at least helps in some cases.
> > 
> > 
> > Perhaps I did not state it very well but it is completely involved in the issue I mentioned as far as I can see. Say you have GPU as default and try to run analysis on the 40 GB molten salt data, RDFs and GK diffusion for example both fail due to memory. This means that our default parameter is not memory safe.
> 
> But doesn't it fail with the current version as well?

It does, and it that sense can be merged, my thought is more that the initial failure here is not deliberate, it has arisen as we see that GPUs will now be used and the efforts made to avoid GPU use did not work as expected. In this case, we would be willingly and knowingly adding code to the project that we know will not be memory safe. I don't think it would be so challenging to do this quickly, let me do some looking and see how we can measure a scaling value accurately and then it should not be a large change.

PythonFZ:
> > > > > I think this is great, however, before we merge it the scaling functions need to be addressed. In my recent attempts at using MDSuite on larger dataset (~100 GB), the scaling functions are not adequate for the smaller memory systems. When we have something like a 90 GB memory system it can absorb some of the failures of the scaling functions but for the GPUs there is no room for this kind of error. Therefore, before making the GPU default, which will almost certainly result in memory failures immediately for those analysing big systems, the scaling functions must be fit correctly.
> > > > 
> > > > 
> > > > I do think, that we can and should merge this anyway. It does not affect the issue you mention, because GPU is default anyways. When a GPU is available TensorFlow will use it, so this at least helps in some cases.
> > > 
> > > 
> > > Perhaps I did not state it very well but it is completely involved in the issue I mentioned as far as I can see. Say you have GPU as default and try to run analysis on the 40 GB molten salt data, RDFs and GK diffusion for example both fail due to memory. This means that our default parameter is not memory safe.
> > 
> > 
> > But doesn't it fail with the current version as well?
> 
> It does, and it that sense can be merged, my thought is more that the initial failure here is not deliberate, it has arisen as we see that GPUs will now be used and the efforts made to avoid GPU use did not work as expected. In this case, we would be willingly and knowingly adding code to the project that we know will not be memory safe. I don't think it would be so challenging to do this quickly, let me do some looking and see how we can measure a scaling value accurately and then it should not be a large change.

I just think that letting this PR go stale for the time being could lead to merge issues later when there is no reason to not merge it.

SamTov:
> > > > > > I think this is great, however, before we merge it the scaling functions need to be addressed. In my recent attempts at using MDSuite on larger dataset (~100 GB), the scaling functions are not adequate for the smaller memory systems. When we have something like a 90 GB memory system it can absorb some of the failures of the scaling functions but for the GPUs there is no room for this kind of error. Therefore, before making the GPU default, which will almost certainly result in memory failures immediately for those analysing big systems, the scaling functions must be fit correctly.
> > > > > 
> > > > > 
> > > > > I do think, that we can and should merge this anyway. It does not affect the issue you mention, because GPU is default anyways. When a GPU is available TensorFlow will use it, so this at least helps in some cases.
> > > > 
> > > > 
> > > > Perhaps I did not state it very well but it is completely involved in the issue I mentioned as far as I can see. Say you have GPU as default and try to run analysis on the 40 GB molten salt data, RDFs and GK diffusion for example both fail due to memory. This means that our default parameter is not memory safe.
> > > 
> > > 
> > > But doesn't it fail with the current version as well?
> > 
> > 
> > It does, and it that sense can be merged, my thought is more that the initial failure here is not deliberate, it has arisen as we see that GPUs will now be used and the efforts made to avoid GPU use did not work as expected. In this case, we would be willingly and knowingly adding code to the project that we know will not be memory safe. I don't think it would be so challenging to do this quickly, let me do some looking and see how we can measure a scaling value accurately and then it should not be a large change.
> 
> I just think that letting this PR go stale for the time being could lead to merge issues later when there is no reason to not merge it.

I appreciate that and have no intention of letting it go stale. I think there is a principle on which to say we should not merge something knowingly that will fail. However, in respect of democracy @christophlohrmann can provide a deciding vote and feel free to merge if he approves.

christophlohrmann:
Do I understand correctly that one cannot turn off the GPU at the moment?

SamTov:
> Do I understand correctly that one cannot turn off the GPU at the moment?

You can turn it off in the same way as here at the moment. The difference is we have a redundant keyword in the calculators. This removes the keyword and defaults to GPU on.

SamTov:
To extend this, I am currently working on the scaling factors in the CI.

christophlohrmann:
if there is a possibility to turn GPU off globally and the memory issues are getting fixed right now, I think the PR is a step in the right direction. If you have concerns about stuff breaking, you can merge this PR into your 'fix scaling factors' PR, otherwise we merge to main now

SamTov:
No concern with breaking, feel free to merge it.

SamTov (approved):
Democratically approved.

PR #475, opened 2022-01-21
Update molecule mapping

SamTov:
The idea in this PR is that I wanted to add an analysis of water as a test and example and in doing so, go through MDSuite functionality and identify bugs and repair them such that the study can continue. I will keep adding small notes on what has changed to aid in the review. It is nothing large just small bug fixes that arise when trying to use the software together. Along with the fixes, I am also writing specific tests to avoid these things from breaking in the future.

#### Summary of additions and changes

Anything Bold-fonted is now included to some extent in a test. To the extent that a test will fail if something that was fixed in this PR is violated in the future.

* **Allowed for building molecule from arbitrary species, i.e. one can now write `{'HW1': 1, 'OW': 1, 'HW2': 1}` and it will build water molecules with these species. This makes it easier to work with GROMACS where typically species are given as a ligand name or simple a numbered species based on its location in a molecule.**
* **Removed molecules from the experiment species dict and updated transformations to look through both species and molecules dicts when performing transformation**
* **Fixed bug in batching for angular distribution function that resulted in errors. This is resolved by setting the default mini batch in the ADF to -1 rather than 50 but I have also added a check in the get_triplets method to completely avoid this issue from occurring.**
* **Adjust ADF such that it can work with molecules**
* **Update transformations so that they work with molecules.**
* Add integration test to rigorously check that the correct water molecules are built from a GROMACS simulation
* Add Functional test to check that MDSuite runs with and without molecules for several calculators
* Adjusted length of some tqdm bars so that they remain at 70 columns
* Added tutorial notebook for mapping molecules.
* **Adapted ADF to work with selected atoms**
* Started adapting RDF to work with selected atoms, see notes below.
* **Fix indexing error for molecule mapping that resulted in a false reference. This was visible, notably, in the centre molecule of the SDF.**
* Fix critical bug in the ensemble generation for remainders and for system-wide properties
* **Move Graph related operations to the graph module**
* **Changed default chemfiles positions property to positions rather that unwrapped positions.**
* Removed indices from molecules, I believe closing #432 
* **Allowed for molecule pbc from gromacs**

#### How to test this pull request

Tests will be added along with a notebook discussing the different methods for building molecules and how they work which can also go into the documentation.

#### Notes

Reviews are welcomed at any point in the PR as well as ideas for improving the mapping module as I go through it. I have some changes planned in terms of its overall functionality so perhaps it is better to wait for that but the choice is yours.

### TODO

[...]

PR #476, opened 2022-01-25
[WIP] SamTov measure memory scaling

### LINKED ISSUE #464: Automatic/standardized scale function computation
Partially in conjunction with #383 but a different consequence, we need a standardized method of assessing the accuracy of the scaling functions. 

My suggestion would be to set up a remote worker with ~4GB of memory, i.e. memory of a standard GPU, and check the scaling functions allow for correct operations on data that is both configuration-wise and atom-wise batched.

SamTov:
Test this PR with 2 GB on [Binder](https://mybinder.org/v2/gh/zincware/MDSuite/SamTov_Measure_Memory_Scaling)

-----

Fix #464 and fix the memory safety of MDSuite which is currently not robust enough for small memory machines (< 8 GB or so) on large data sets (>10 GB). 

### Memory scaling measurement methodology

It seems one can use the pytest-monitoring pytest plugin to measure the memory usage of specific tests. The outputs of these measurements are stored in an sqlite database that can then be queried for information about the specific test that was run. The workflow is outlined below:

1. Turn off memory management in a calculator
2. Add several experiments to a project, each with a different size of data in steps [1, 10, 50, 100] MBs. In the case of species wise computations different experiments are not needed however for flux data we cannot enforce computations over different groups and therefore, several experiments will be required.
3. Run the calculator for different databases and measure memory scaling w.r.t the dominant scaling axis
    - In the case of an atom-wise calculators this may be data range or number of atoms depending on the operation.
4. Fit a function to the relationship and try to evaluate the scaling function exactly.

This requires the person running the experiment to install **`pytest-monitoring`** locally. It can be added theoretically to the CI on GitHub and be tested every time but we will then have to make the SQL reading more robust. 

### Tasks  

- [x] Add general framework for memory testing using pytest monitor
- [ ] Add functionality for adding a hdf5 database directly into an MDSuite experiment
- [x] Add functionality for turning off memory management in case of a test (all batch sizes to 1 and no minibatching)
- [ ] Add curve fitting to the test for identify correct relationship.
- [x] Exclude the memory scaling test from the CI
- [ ] Adjust GitHub runners to use 4 GB memory machine to ensure that memory safety is achieved on a reasonably memory limitd device.

### Reviewer notes
Any assistance with this is welcomed as it is not the only important PR on MDSuite at the moment, namely, #475 and this still need to be fixed before any release.


PythonFZ:
I would have some questions:
> Add functionality for adding a hdf5 database directly into an MDSuite experiment

What do you mean by that?

> Exclude the memory scaling test from the CI

I did this for MLSuite with `@pytest.mark.gap` and then run `pytest -m "not gap"` and I think it makes sense to go with the same framework for MDSuite, but we could also name the files differently, idk.

> Adjust GitHub runners to use 4 GB memory machine to ensure that memory safety is achieved on a reasonably memory limitd device.

The do have 7 GB of memory for linux / windows and I dont't think we can change that.

CI/integration_tests/calculators/test_radial_distribution_function.py
```diff
@@ -30,6 +30,7 @@
 from zinchub import DataHub
 
 import mdsuite as mds
+mds.config.memory_fraction = 1
 from mdsuite.utils.testing import assertDeepAlmostEqual
```
PythonFZ: We have to be careful here, where to set that! If you set it here if will be used in all tests iirc. So it should go in the test definition or in a pytest.fixture.

mdsuite/utils/config.py
```diff
@@ -36,10 +38,22 @@ class Config:
...
+            are set to 1. Should typically be accompanied by the memory fraction being
+            set to 1 as well.
+    memory_fraction: bool
+            The portion of the available memory to be used.
     """
 
     jupyter: bool = False
     GPU: bool = False
+    memory_scaling_test: bool = False
+    memory_fraction: float = 0.5
```
PythonFZ, line 56: Maybe we could expand the config to be `config.memory.scaling_test = True` instead of `config.memory_scaling_test = True` with additional dataclasses. This way it could be more structured.

CI/integration_tests/calculators/test_radial_distribution_function.py
SamTov: Yes sorry it shouldn't be here at all. This was something I used to check if it was working and not at al meant to be in the PR.

mdsuite/utils/config.py
SamTov: Yeah that's an interesting idea. Managing how configuration things are set in general is a nice thing to discuss as it can be quite involved. I think having data classes for different things like you mention here would be very nice.

CI/integration_tests/calculators/test_radial_distribution_function.py
SamTov: In general in the actual running tests I will not have any `scaling=True` unless we want it to be done automatically each time. This is a general thing we can discuss but I am not sure it is necessary.

SamTov:
> I would have some questions:
> 
> > Add functionality for adding a hdf5 database directly into an MDSuite experiment
> 
> What do you mean by that?

So what I want to do, and what I have started in the test module, is to generate several hdf5 databases with data of different exact sizes, e.g 1MB, 10MB and so on. In this case, rather than generate a numpy array, save it to a readable file, and then read it with MDSuite, I want to make hdf5 database, make an experiment, and then add it as data to that experiment.

In practice, this would be the equivalent of saving your simulation data into the H5MD database format and then using it in MDSuite.

> > Exclude the memory scaling test from the CI
> 
> I did this for MLSuite with `@pytest.mark.gap` and then run `pytest -m "not gap"` and I think it makes sense to go with the same framework for MDSuite, but we could also name the files differently, idk.
So the only factor here is whether we want to have it run each time? I think the it is unnecessary to have the scaling stuff run every time but am open to differing opinions.

> > Adjust GitHub runners to use 4 GB memory machine to ensure that memory safety is achieved on a reasonably memory limitd device.
> 
> The do have 7 GB of memory for linux / windows and I dont't think we can change that.

I thought somewhere you can set a memory limit but maybe I am mistaken. In that case, we can set up local runners and just keep it as a test for say, releases that we perform locally.

CI/integration_tests/calculators/test_radial_distribution_function.py
SamTov: This has been removed.

PR #484, opened 2022-01-28
fix #482

### LINKED ISSUE #482: Check log Level for log file
I just realized that the logger does neither print log.debug() lines nor store them in the mdsuite.log file. 
where can you even see log.debug() lines then?

_Originally posted by @MarcBrueck in https://github.com/zincware/MDSuite/pull/478#discussion_r794290759_

PythonFZ:
Fix #482

### DIFF
--- a/mdsuite/__init__.py
+++ b/mdsuite/__init__.py
@@ -37,7 +37,7 @@ __all__ = ["Project", "Experiment", "adjacency_matrix", "Report", "config"]
 __version__ = "0.1.0"
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


SamTov (approved):
I found initially when this was set that it would aways print the DEBUG messages. But if this is not the case then its okay.

PythonFZ:
> I found initially when this was set that it would aways print the DEBUG messages. But if this is not the case then its okay.

There is another log level set for the formatter which remains INFO for std-out and DEBUG for the file log.

PR #485, opened 2022-01-30
Sam tov post rdf calculators

### LINKED ISSUE #327: CN, StructureFactor, POMF should have the same args as RDF
Because they rely on the RDF calctulation one should be able to pass the parameters for the RDF.

This is related to #323 

Also replace `_get_rdf_data` with `exp.load.RDF()`

### LINKED ISSUE #479: fix `_ang_to_nm` in RDF calculator
I should note here that the name of the method is very misleading, it does not convert angstroms to nm it converts the simulation unit 'length' to nm. I will update this.

Method is shown below for clarity:

```python
def _ang_to_nm(self, data_in: np.ndarray) -> np.ndarray:
        """
        Convert Angstroms to nm
        Returns
        -------
        data_out : np.ndarray
                data_in converted to nm
        """
        return (self.experiment.units["length"] / 1e-9) * data_in
```

Metal units are Angstroms not in nm so in this case, when the radial values come back as 1xe-11 it makes sense to me. It does not matter what units you run your simulation in, the RDF will be saved in nm.

_Originally posted by @SamTov in https://github.com/zincware/MDSuite/issues/433#issuecomment-1023009666_

### LINKED ISSUE #481: Passing arguments post-operation calculators
We should decide on how to pass arguments to calculators which rely on the computation of another calculator. At the moment this is typically only RDFs but we also have Nernst-Einstein computations which could use such a feature. 

Currently, there is no parameter information passed, the calculator just tries to load an arbitrary RDF from the SQL database and runs one if none exists. This means that if you computed 3 RDFs under different conditions and want to look at the coordination numbers of all of them, you cannot. This becomes quite important if you study molecules as you may want to look at the atomistic coordination numbers or KB integrals as well as those of the molecules which is not possible.

My suggestion is as follows:

```python
rdf_data = experiment.run.RadialDistributionFunction()

cn_data = experiment.run.CoordinationNumbers(rdf_data, savgol_filter_parameters, ...)
```
Or for Nernst Einstein 
```python
self_diffusion_data = experiment.run.EinsteinDiffusion()
correlated_diffusion_data = experiment.run.GreenKuboDistinctDiffusion()

ne_cond = experiment.run.NersnstEinstein(
        self_diffusion = self_diffusion_data, correlated_diffusion = correlated_diffusion_data
)
```

In cases where defaults can be had, i.e those relying on RDFs, we can set the default in the calculator to call an RDF computation with the default parameters and proceed. 

Any and all comments and thoughts are welcomed.
@Fratorhe @PythonFZ @christophlohrmann @MarcBrueck

SamTov:
[Binder](https://mybinder.org/v2/gh/zincware/MDSuite/SamTov_PostRDF_Calculators)

Add functionality for passing RDFs to post-rdf calculators and fix up post-RDF calc that do not work.

### Changes
* Fixed failures in CN calculator.
* Allowed for N coordination shells in CN calculator
* Allowed for parsed data to post-RDF calculators.
* Allow for N pomf shell computations in POMF
* Add comparison notebook to highlight features.
* Enable RDF filtering in KB integrals as exists in POMF and CN
* Inherit computation parameters from parsed RDF
* Implement new SF code.
* Add memory management requires tests with publication backed data for coordination numbers


SamTov:
Aside from adding tests and examples and fixing the SF calculator I do not see many updates coming here and will probably leave it for a day or two so perhaps a review now could be efficient.

examples/scripts/structure_factor_tutorial.py
```diff
@@ -0,0 +1,21 @@
...
+"""Make sure to give your particles the right charge,
+otherwise the calculated structure factor will be incorrect"""
+project.experiments["NaCl_1399K"].set_charge("Na", [+1])
+project.experiments["NaCl_1399K"].set_charge("Cl", [-1])
+
+# You also need to use the correct mass, but this is usually already set
+# project.experiments['NaCl_1399K'].set_mass('Cl',[50])
+
+
+project.run.StructureFactor(plot=True)
```
PythonFZ: Do we have some data that we can use for this test?

mdsuite/calculators/coordination_number_calculation.py
```diff
@@ -109,267 +140,279 @@ def __init__(self, **kwargs):
...
         """
+        Compute the volume of the box in nm.
 
-        species = species.split("_")  # get an array of the species being studied
-        rdf_number_of_atoms = self.experiment.species[species[0]].n_particles
+        Returns
+        -------
+        Updates the volume attribute of the class.
+        """
+        volume_si = self.experiment.volume * (self.experiment.units["length"] ** 3)
```
PythonFZ, line 218: Not related to the PR but we could consider looking more into [Pint](https://pint.readthedocs.io/en/stable/) for our unit systems.

mdsuite/calculators/coordination_number_calculation.py
```diff
@@ -109,267 +140,279 @@ def __init__(self, **kwargs):
...
-            ]
-        )
+        coordination_shells = {}
+        for i in range(self.args.number_of_shells):
+            coordination_shells[i] = np.zeros(2, dtype=int)
+            cn_radii_range = golden_section_search(
+                [radii, rdf], radii[peaks[i + 1]], radii[peaks[i]]
+            )
+            for j in range(2):
+                coordination_shells[i][j] = np.where(radii == cn_radii_range[j])[0][0]
```
PythonFZ, line 299: I guess `np.squeeze(np.where(radii==cn_radii_range[j]))` would also work - is it a single value? But also no need to change that.

mdsuite/calculators/coordination_number_calculation.py
```diff
@@ -109,267 +140,279 @@ def __init__(self, **kwargs):
...
-                    self.integral_data[self.indices[0][0]],
-                    self.integral_data[self.indices[0][1]],
-                ]
+        coordination_numbers = {}
+
+        for key, val in coordination_shells.items():
+            lower_bound = integral_data[val[0]]
+            upper_bound = integral_data[val[1]]
+
+            coordination_numbers[f"CN_{int(key) + 1}"] = np.mean(
```
PythonFZ, line 332: maybe add a function `get_key` `lamnda key: f"CN_{int(key) + 1}"` so we could potentially change how it is stored in a single place.

mdsuite/calculators/coordination_number_calculation.py
```diff
@@ -109,267 +140,279 @@ def __init__(self, **kwargs):
...
+        for selected_species, vals in self.rdf_data.data_dict.items():
             log.debug(f"Computing coordination numbers for {selected_species}")
-            self.radii = np.array(vals["x"]).astype(float)[1:]
-            self.rdf = np.array(vals["y"]).astype(float)[1:]
-            self.selected_species = selected_species.split("_")
-            self.species_tuple = selected_species  # depreciated
 
-            density = self._get_density(self.selected_species[0])  # calculate the density
+            radii = np.array(vals["x"]).astype(float)[1:]
+            rdf = np.array(vals["y"]).astype(float)[1:]
```
PythonFZ, line 350: This should use `RadialDistributionFunction.series_keys` so that we don't have to rely on it being always `x` and `y`.
Maybe it would be best if we added the keys as class attributes so that they can be read from the class without init?

mdsuite/calculators/coordination_number_calculation.py
```diff
@@ -109,267 +140,279 @@ def __init__(self, **kwargs):
...
-            model_2 = BoxAnnotation(
-                left=val[self.result_keys[4]],
-                right=val[self.result_keys[5]],
-                fill_alpha=0.1,
-                fill_color="red",
+            fig = figure(x_axis_label=self.x_label, y_axis_label=self.y_label)
+
+            # Add vertical lines to the plot
+            for i in range(self.args.number_of_shells):
+                coordination_number = val[f"CN_{i + 1}"]
```
PythonFZ, line 379: here the function could apply again

mdsuite/calculators/potential_of_mean_force.py
```diff
@@ -126,157 +132,253 @@ def __init__(self, **kwargs):
...
+                "number_of_configurations"
+            ],
+            number_of_shells=number_of_shells,
         )
 
-    def _calculate_potential_of_mean_force(self):
+        # Auto-populate the results.
+        for i in range(self.args.number_of_shells):
+            self.result_keys.append(f"POMF_{i + 1}")
+            self.result_keys.append(f"POMF_{i + 1}_error")
```
PythonFZ, line 181: could also benefit from a common function

mdsuite/calculators/structure_factor.py
```diff
@@ -99,246 +110,264 @@ class StructureFactor(Calculator):
...
+        # set args that will affect the computation result
+        self.args = Args(
+            number_of_bins=self.rdf_data.computation_parameter["number_of_bins"],
+            cutoff=self.rdf_data.computation_parameter["cutoff"],
+            number_of_configurations=self.rdf_data.computation_parameter[
+                "number_of_configurations"
+            ],
+            resolution=resolution,
+        )
+        self.q_values = np.linspace(0.5, 12, resolution)
```
PythonFZ, line 180: Out of curiosity - why is it from `0.5 -> 12`

mdsuite/calculators/structure_factor.py
```diff
@@ -99,246 +110,264 @@ class StructureFactor(Calculator):
...
 
-    def atomic_form_factors(self, scattering_scalar):
+        Returns
+        -------
+        Updates the data class for each species.
         """
-        Calculates the atomic form factors for all elements in the species
-        dictionary and returns it
+        for name, species_data in self.species_dict.items():
+            aff_data = self.form_factor_data.loc[self.form_factor_data["Element"] == name]
```
PythonFZ, line 230: maybe change `aff_data` to something more descriptive. But I don't know if it is already descriptive for most people it would be fine.

mdsuite/calculators/structure_factor.py
```diff
@@ -99,246 +110,264 @@ class StructureFactor(Calculator):
...
-                * self.experiment.box_array[2]
-            )
-        return species_densities
+        structure_factor = np.zeros(self.args.resolution)
+        for pair, pair_data in partial_sf.items():
+            species = pair.split("_")
+            if species[0] == species[1]:
+                factor = 2
+            else:
+                factor = 1
```
PythonFZ, line 321: Confuse people with:
`factor = int(species[0] == species[1]) + 1`

PythonFZ (commented):
Looks very good so far - some tests with different RDF data to ensure that it works should be added but because WIP it looks good.

examples/scripts/structure_factor_tutorial.py
SamTov: I will add tests later and I have data for all of them

mdsuite/calculators/coordination_number_calculation.py
SamTov: Perhaps a good talking point at the assembly

mdsuite/calculators/coordination_number_calculation.py
SamTov: I think in that case they would have to come from the data dict provided as the CNs have their own series keys.

mdsuite/calculators/structure_factor.py
SamTov: It is set to 12 for me to look at a plot nicely, in general it will be set to 25 as that is range over which the approximation of Gaussian basis functions on the form factors is valid. Not set to 0 to avoid an explosion.

mdsuite/calculators/structure_factor.py
SamTov: I love it

PythonFZ: @christophlohrmann suggested `species = 2 if species[0] == species[1] else 1`. I don't understand why anyone could dislike casting boolean to integer.

PythonFZ:
For the Notebook I would like to change all occurrences of `data.data_dict['<some_key>']` to `data['<some_key>']`. The `data_dict` option will remain but I prefer the - in my opinion - cleaner direct `getitem` call.

mdsuite/calculators/structure_factor.py
SamTov: In general I think keeping one-liners that aren't extremely transparent out of code. I dont always think the shortes solution is the best when it comes to code clarity, particularly is performance is not impacted.

SamTov:
> For the Notebook I would like to change all occurrences of `data.data_dict['<some_key>']` to `data['<some_key>']`. The `data_dict` option will remain but I prefer the - in my opinion - cleaner direct `getitem` call.

That I am in the process of doing for all of the examples.

mdsuite/calculators/structure_factor.py
SamTov: For this one, in another method I wrote it in the doc string as there is not reasonable way of writing `atomic_form_factor_data` with such a long assignment. My though here for additional clarity would be to add it to the doc-string and add a comment over its first assignment along the lines of:

```python
# aff -> atomic form factor
aff_data = ...
```

PythonFZ: sounds good

mdsuite/calculators/potential_of_mean_force.py
SamTov: I think when the calculators are restructured the way that recorded data is described in a calculator will be reassessed completely. For now I would leave this and change it when we change all of them.

CI/functional_tests/test_molten_salts.py
```diff
@@ -90,18 +95,59 @@ def mdsuite_project(traj_files, tmp_path) -> mds.Project:
...
     """
-    pass
+    NaCl_experiment = mdsuite_project.experiments.NaCl
+    KCl_experiment = mdsuite_project.experiments.KCl
+
+    RDF_Data = mdsuite_project.run.RadialDistributionFunction(
+        number_of_configurations=500, cutoff=15.0, plot=False
+    )
+    NaCl_CN_data = NaCl_experiment.run.CoordinationNumbers(
+        rdf_data=RDF_Data["NaCl"],
```
PythonFZ, line 127: I do think the nicer way to test these would be `NaCl_RDF_data = NaCl_experiment.run.RDF` and then pass `rdf_data = NaCl_RDF_data` without the dict. For the tests it does not make a difference I suppose.

CI/functional_tests/test_molten_salts.py
```diff
@@ -90,18 +95,59 @@ def mdsuite_project(traj_files, tmp_path) -> mds.Project:
...
+    NaCl_CN_data["Na_Cl"]["CN_1"] == pytest.approx(5.213, 0.0001)
+    NaCl_CN_data["Na_Cl"]["CN_2"] == pytest.approx(35.090, 0.0001)
+    NaCl_CN_data["Na_Na"]["CN_1"] == pytest.approx(14.775, 0.0001)
+
+    KCl_CN_data["Cl_K"]["CN_1"] == pytest.approx(5.507, 0.0001)
+
+    KCl_POMF_data["Cl_K"]["POMF_1"] == pytest.approx(-1.372e-11, 1e-14)
+
+    mdsuite_project.run.GreenKuboDiffusionCoefficients()
+    mdsuite_project.run.EinsteinDiffusionCoefficients()
```
PythonFZ, line 157: Do we have data for these?

CI/integration_tests/calculators/__test_structure_factor.py
```diff
@@ -62,7 +62,7 @@ def test_structure_factor_project(traj_files, true_values, tmp_path):
     )
 
     project.run.RadialDistributionFunction(number_of_configurations=-1, plot=False)
-    project.run.StructureFactor(plot=False)
+    # project.run.StructureFactor(plot=False)
```
PythonFZ: The test should be disabled, so this comment should not be required?

CI/integration_tests/calculators/test_coordination_numbers.py
```diff
@@ -51,21 +51,18 @@ def true_values() -> dict:
...
+        "NaCl", simulation_data=traj_file, timestep=0.002, temperature=1400
     )
 
-    computation = project.run.CoordinationNumbers()
+    computation = project.run.CoordinationNumbers(
+        savgol_order=3, savgol_window_length=111, plot=False
+    )
 
-    keys = project.run.CoordinationNumbers.result_keys
+    data_dict = computation["NaCl"].data_dict["Na_Cl"]
```
PythonFZ: should be `computation["NaCl"]["Na_Cl"]` - on that note, we should consider tuples (#490 )

CI/integration_tests/calculators/test_green_kubo_self_diffusion_coefficients.py
```diff
@@ -61,4 +61,8 @@ def test_project(traj_file, true_values, tmp_path):
 
     computation = project.run.GreenKuboDiffusionCoefficients(plot=False)
 
+    for key, val in computation["NaCl"].data_dict.items():
```
PythonFZ: I think `computation["NaCl"].items()` should also work - I have to check that though

PythonFZ: also, why iterate over `items` and not `values`?

examples/scripts/structure_factor_tutorial.py
```diff
@@ -0,0 +1,21 @@
+import mdsuite as mds
+
```
PythonFZ: Is this file still relevant?

mdsuite/calculators/__init__.py
```diff
@@ -50,7 +50,7 @@
 from .radial_distribution_function import RadialDistributionFunction
```
PythonFZ: I think absolute imports are favourable

mdsuite/calculators/einstein_distinct_diffusion_coefficients.py
```diff
@@ -100,7 +100,7 @@ def __init__(self, **kwargs):
         """
         super().__init__(**kwargs)
 
-        self.scale_function = {"linear": {"scale_factor": 10}}
+        self.scale_function = {"quadratic": {"inner_scale_factor": 10}}
```
PythonFZ, line 103: I don't remember exactly when and where we already mentioned that, but these should be dataclasses / enums and not strings - maybe we should open an issue for that?

mdsuite/calculators/potential_of_mean_force.py
```diff
@@ -126,157 +132,253 @@ def __init__(self, **kwargs):
...
+        Use the provided RDF data to populate the args class.
+
+        Returns
+        -------
+        number_of_bins : int
+                The data range used in the RDF calculation.
+        cutoff : float
+                The cutoff (in nm) used in the RDF calculation
+        """
+        raw_data = self.rdf_data.data_dict
```
PythonFZ, line 217: should not use `data_dict`

mdsuite/calculators/potential_of_mean_force.py
```diff
@@ -126,157 +132,253 @@ def __init__(self, **kwargs):
...
 
-        return [peaks[0], peaks[1]]
+        # Check that the required number of peaks exist.
+        if len(peaks) < required_peaks:
+            msg = (
+                "Not enough peaks were detecting in the RDF to perform the desired "
+                "analysis. Try reducing the number of desired peaks or improving the "
+                "quality of the RDF provided."
+            )
+            log.error(msg)
```
PythonFZ, line 261: More a general question - do we need logging if we raise an error? And If we do this should be critical, not error iirc.

mdsuite/experiment/run.py
```diff
@@ -13,8 +13,10 @@
 import functools
 from typing import TYPE_CHECKING, Any, List, Type, Union
 
-from mdsuite.calculators import (  # SpatialDistributionFunction,
-    AngularDistributionFunction,
+from mdsuite.calculators import (
+    AngularDistributionFunction,  # SpatialDistributionFunction,
+)
+from mdsuite.calculators import (  # StructureFactor,
```
PythonFZ, line 19: How did this pass isort?

PythonFZ (changes requested):
a few changes for the notebooks I'd like to leave  here
- [ ] change `savgol_window_length=111` in the notebooks
- [ ] remove comments from the notebooks

CI/functional_tests/test_molten_salts.py
SamTov: Yeah I think for most of the tutorials I split this up but here I would not bother too much.

CI/functional_tests/test_molten_salts.py
SamTov: In principle I can add data but as was suggested in the previous tests I can only provide some references of papers with approximate values. Testing the complete diffusion value and msd is done in the proper LJ verified test. In case of GK we have not a proper test yet. 

A benefit of this test though is that it has to be batched and therefore tests some of the memory management and the ensembling.

CI/integration_tests/calculators/__test_structure_factor.py
SamTov: resolved.

CI/integration_tests/calculators/test_green_kubo_self_diffusion_coefficients.py
SamTov: I usually use both just in case I need more information but here it is not important. This test I only changed as it needed to have some additional information removed from it. I can remove the data dict call though.

examples/scripts/structure_factor_tutorial.py
SamTov: no I will remove it

mdsuite/calculators/__init__.py
SamTov: I think they are essential

mdsuite/calculators/einstein_distinct_diffusion_coefficients.py
SamTov: I think this can all be handled in the calculator restructure as it is not clear whether these will exist or not after that.

mdsuite/calculators/potential_of_mean_force.py
SamTov: I think logging errors is good in any case as it gets stored somewhere together with the other things that took place. As for the type of logging I am not the person to ask.

mdsuite/experiment/run.py
SamTov: I guess isort will have to answer that

SamTov:
@christophlohrmann & @PythonFZ  I am relatively confident the kinks have been worked out on this PR but needs to be updated to the new data configurations in the CI. I think this should be a priority.

PythonFZ:
> @christophlohrmann & @PythonFZ I am relatively confident the kinks have been worked out on this PR but needs to be updated to the new data configurations in the CI. I think this should be a priority.

You can cherry-pick or merge with main. But the pytest build / nbtest seem to fail for other reasons. If the tests are fixed I agree that this can be merged soon.

SamTov:
> > @christophlohrmann & @PythonFZ I am relatively confident the kinks have been worked out on this PR but needs to be updated to the new data configurations in the CI. I think this should be a priority.
> 
> You can cherry-pick or merge with main. But the pytest build / nbtest seem to fail for other reasons. If the tests are fixed I agree that this can be merged soon.

nbtest uses datahub so it will fail. If it has no data I can image debugging will be difficult. In the case of the pomf and cn it will be a different tag and as you were responsible for that PR I would appreciate that you and @christophlohrmann take care of that.

PythonFZ (approved):
I  guess if the tests pass this PR is ready

christophlohrmann (approved):
I do not pretend to understand the changes but since the tests pass I trust it

PR #493, opened 2022-02-02
Test and fix msd

### LINKED ISSUE #492: MSD wrong
The prefactor ``6`` is applied to the msd, not just to the diffusion coefficient, so all MSDs are off by a factor of 6

Note: PR is on the way

christophlohrmann:
[Binder](https://mybinder.org/v2/gh/zincware/MDSuite/test_msd)

Fix #492 

One by one we should replace all data-driven tests by tests with controlled input


christophlohrmann:
we can either fix the water test with reliable 'truth values' or remove value checking there

mdsuite/file_io/chemfiles_read.py
```diff
@@ -103,7 +103,7 @@ def _get_metadata(self) -> TrajectoryMetadata:
                 mdsuite.database.simulation_database.SpeciesInfo(
                     name=key,
                     n_particles=len(val),
-                    properties=list(self.properties_to_chemfile_attr_dict.keys()),
+                    properties=list(self.properties_in_file.keys()),
```
PythonFZ, line 106: See https://google.github.io/styleguide/pyguide.html#284-decision

SamTov (approved):
It looks good, I may update the water tests, particularly if the test fails but I think it is nice. One thing we can consider about the data is that the water is studied using SPCE. This means that the molecular structure of each water molecule is a validated quantity, i.e. the ADF of one molecule is validated by the SPCE parametrization. Therefore, what we can use in a formal test is group ADFs and theoretically group RDFs when they work.

SamTov (changes requested):
Something is broken with the time computation in this branch, will request changes to ensure nothing can be merged for now.

SamTov (commented):
bokeh_pic

The times for each of the species here are different but should be the same. This is because looping over the species takes place within the calculator class and therefore, the time attribute, not having been reset, is multiplied again by the sim-SI scaling. The other thing here is that the time is orders of magnitude too small. This simulation is working on ns scales.

I am moving the time and msd instantiation under the species loop to avoid this.

**Status**
Resolved

christophlohrmann:
@SamTov what is the status here?

SamTov:
> @SamTov what is the status here?

@christophlohrmann  The diffusion coefficients don't match those of Green-Kubo or those taken from a LAMMPS MSD computation. I haven't tried in the recent days to look into why.

SamTov:
@christophlohrmann I have found that the computed MSDs and times appear to be the same before and after the change with the obvious x6 scaling, so that is one factor removed. I did this by comparing the MSDs between two branches in the example, the coefficients differ widly (by a factor for 100 for this data actually) which suggests something to do with the curve fitting perhaps.


**UPDATE**

Manually computing the diffusion coefficient on the data generator yields a value similar to that of the previous calculations. Something appears to be different with how the diffusion coefficient is fit.

SamTov:
Currently it looks like the following:

Screen Shot 2022-02-23 at 20 56 42

The diffusion coefficient line looks very wavy but if you check the dimensions on the axis it will explain why. The range is set between 0.999 the min value and 1.001 the max.

PythonFZ:
> Currently it looks like the following:
> 
> Screen Shot 2022-02-23 at 20 56 42
> 
> The diffusion coefficient line looks very wavy but if you check the dimensions on the axis it will explain why. The range is set between 0.999 the min value and 1.001 the max.

I like the fit line and the error on the plot. Maybe we could scale the y axis a bit? I know we would loose some information but this way it might add confusion? Do you think the line is useful in general? I'm definetly for keeping it, but maybe through an `plot_diffusion_over_time=False` to enable it manually?

SamTov:
> > Currently it looks like the following:
> > Screen Shot 2022-02-23 at 20 56 42
> > The diffusion coefficient line looks very wavy but if you check the dimensions on the axis it will explain why. The range is set between 0.999 the min value and 1.001 the max.
> 
> I like the fit line and the error on the plot. Maybe we could scale the y axis a bit? I know we would loose some information but this way it might add confusion? Do you think the line is useful in general? I'm definetly for keeping it, but maybe through an `plot_diffusion_over_time=False` to enable it manually?

Yeah I actually think it is only useful if it is very wrong, i.e, that y range is much bigger than expected. We can have it turned off quite easily that is no problem. Also keep in mind that we have interactive zooming on the plots so we can also just have it quite a bit smaller on the normal axis so that it looks straight but then users can always zoom in to see how it looks up close.

christophlohrmann:
What is on the x-axis for the diffusion coefficient plot?
Also is there a way to see which original values went into the calculation of the final diffusion coefficient?

SamTov:
> What is on the x-axis for the diffusion coefficient plot? Also is there a way to see which original values went into the calculation of the final diffusion coefficient?

The x axis is time still, it just starts at the first point in which the msd was taking, we could also add the start line to the plot as well which would show to full range over which it was fit.

That data is stored in the SQL database now yes.

christophlohrmann:
so convergence in the plot goes from right to left?

SamTov:
> so convergence in the plot goes from right to left?

Nope it goes left to right,the line symbolizes the end point of a fit so the last one means he fit was taken all the way to the last point.

christophlohrmann:
ok so start point gets picked automatically and is constant, and then the end point varies. In that case: approved

SamTov:
> ok so start point gets picked automatically and is constant, and then the end point varies. In that case: approved

Exactly. I will fix up the tests and try to resolve the plot labels first.

PythonFZ:
Can we mark the starting point of the fit with either a vertical line or what I would probably prefer a small `x` on the plot? I think that is a bit easier to see than the starting point of the orange line?

CI/functional_tests/test_water_study.py
```diff
@@ -79,13 +79,14 @@ def mdsuite_project(traj_files, tmp_path) -> mdsuite.Project:
...
 
-    project.add_experiment(
+    exp = project.add_experiment(
         name="water_sim",
         timestep=0.002,
         temperature=300.0,
         units=gmx_units,
         simulation_data=file_reader,
     )
+    exp.sample_rate = 5000
```
PythonFZ, line 91: Not for this PR, but we could consider having an `add_experiment(..., **kwargs)` and then do `[setattr(self, key, value) for key, value in kwargs.items]` so we can pass missing information directly through this method and don't have to call `exp.sample_rate` ?

mdsuite/calculators/einstein_diffusion_coefficients.py
```diff
@@ -112,6 +121,7 @@ def __call__(
         atom_selection: np.s_ = np.s_[:],
         molecules: bool = False,
         tau_values: Union[int, List, Any] = np.s_[:],
+        fit_range: int = -1,
```
PythonFZ, line 124: I would vote for `None` over `-1`

SamTov: The -1 here was sort of motivated by the fact that array[-1] is the last element which is where it is taken.

PythonFZ: Okay, then I like the `-1`. But maybe the name `range` is a bit misleading then? So this defines the stop/max value of the range argument? Maybe name it `fit_max` / `fit_stop` or so? But given that `for x in range(stop)` automatically is the stop value we can also keep it.

SamTov: Yeah I also struggled with the name here, it isn't a range without the starting point it is just a stop index. fit max index or somethign maybe?

PythonFZ: Don't we have someone in our team who gave a talk on variable naming @christophlohrmann ?

christophlohrmann: fit_max_index is better than range imo. It also tells that an int is expected (unlike tau_values)

SamTov: sounds good to me.

SamTov:
> Can we mark the starting point of the fit with either a vertical line or what I would probably prefer a small `x` on the plot? I think that is a bit easier to see than the starting point of the orange line?

Should I then make all of the points markers and not line spans? I can mark the user defined stop index, the start index, and the end of the data by markers. I also played with the idea of shading the area over which the fit was taken.

christophlohrmann:
> > Can we mark the starting point of the fit with either a vertical line or what I would probably prefer a small `x` on the plot? I think that is a bit easier to see than the starting point of the orange line?
> 
> Should I then make all of the points markers and not line spans? I can mark the user defined stop index, the start index, and the end of the data by markers. I also played with the idea of shading the area over which the fit was taken.

shading the final fit area would be really nice

SamTov:
> > > Can we mark the starting point of the fit with either a vertical line or what I would probably prefer a small `x` on the plot? I think that is a bit easier to see than the starting point of the orange line?
> > 
> > 
> > Should I then make all of the points markers and not line spans? I can mark the user defined stop index, the start index, and the end of the data by markers. I also played with the idea of shading the area over which the fit was taken.
> 
> shading the final fit area would be really nice

That should not be a problem.

SamTov:
> > What is on the x-axis for the diffusion coefficient plot? Also is there a way to see which original values went into the calculation of the final diffusion coefficient?
> 
> The x axis is time still, it just starts at the first point in which the msd was taking, we could also add the start line to the plot as well which would show to full range over which it was fit.
> 
> That data is stored in the SQL database now yes.

@christophlohrmann Just to confirm here, there is no averaging over diffusion coefficients to get a final value in this approach. Only the diffusion coefficient computed for the user defined fit_end (last point by default) is returned as the official coefficient and the error is simply the fit error. The other gradients are included for reference and are stored as gradients and gradient errors in the database in case someone see that an earlier point would have been a better fit option.

christophlohrmann:
The water test values have changed now about 5 times. How do we know if any of them will be finally correct?

SamTov:
> The water test values have changed now about 5 times. How do we know if any of them will be finally correct?

I can remove the value check for diffusion, for structure stuff we can match experiments as it is the SPCE model. The coefficients are now all on the correct scale and if we know that the real diffusion test is in there and passes then this diffusion value should also be correct. Visual inspection of the msds also shows very good fitting and the simulation parameters like time step and sample rate are confirmed correct now.

In addition, the water test that is subject to small random fluctuations, the single group diffusion, has had its assert statement removed in favour of just checking for a successful run.

christophlohrmann:
> > The water test values have changed now about 5 times. How do we know if any of them will be finally correct?
> 
> I can remove the value check for diffusion, for structure stuff we can match experiments as it is the SPCE model. The coefficients are now all on the correct scale and if we know that the real diffusion test is in there and passes then this diffusion value should also be correct. Visual inspection of the msds also shows very good fitting and the simulation parameters like time step and sample rate are confirmed correct now.
> 
> In addition, the water test that is subject to small random fluctuations, the single group diffusion, has had its assert statement removed in favour of just checking for a successful run.

With heavy heart I vote for removing the value check since it is not obvious what the value should be

SamTov:
I would like to handle all plotting in an alternative PR as it should be adjusted for many files not just the ones here but all Einstein and GK plots. However, for this PR, should the changes made to how the data is stored, i.e, msds not with the /6 be propagated to all Einstein computations? I think this is within the scope of the PR and should be done before merging. @PythonFZ @christophlohrmann

SamTov:
> > > The water test values have changed now about 5 times. How do we know if any of them will be finally correct?
> > 
> > 
> > I can remove the value check for diffusion, for structure stuff we can match experiments as it is the SPCE model. The coefficients are now all on the correct scale and if we know that the real diffusion test is in there and passes then this diffusion value should also be correct. Visual inspection of the msds also shows very good fitting and the simulation parameters like time step and sample rate are confirmed correct now.
> > In addition, the water test that is subject to small random fluctuations, the single group diffusion, has had its assert statement removed in favour of just checking for a successful run.
> 
> With heavy heart I vote for removing the value check since it is not obvious what the value should be

I can agree with that. I will do this update now.

christophlohrmann:
> I would like to handle all plotting in an alternative PR as it should be adjusted for many files not just the ones here but all Einstein and GK plots. However, for this PR, should the changes made to how the data is stored, i.e, msds not with the /6 be propagated to all Einstein computations? I think this is within the scope of the PR and should be done before merging. @PythonFZ @christophlohrmann

I have no problem with a partial rollout of the plotting, bit by bit we can adapt all calculators.
As to the 6: If you know that the other MSDs are wrong, we should correct them as fast as possible

SamTov:
> > I would like to handle all plotting in an alternative PR as it should be adjusted for many files not just the ones here but all Einstein and GK plots. However, for this PR, should the changes made to how the data is stored, i.e, msds not with the /6 be propagated to all Einstein computations? I think this is within the scope of the PR and should be done before merging. @PythonFZ @christophlohrmann
> 
> I have no problem with a partial rollout of the plotting, bit by bit we can adapt all calculators. As to the 6: If you know that the other MSDs are wrong, we should correct them as fast as possible

There are 5 Einstein calculators and there will likely be a /6 in all of them with exception now of course of the self-diffusion. This is probably also true for GK and a /3

SamTov:
@PythonFZ The tests fail now due to a type error with the new Molecules dict api, could you handle it as you likely have the fastest idea of that api and how it works with actual molecule names?

```bash
>       water_group = water.molecules["water"]["groups"]["0"]
E       TypeError: 'MoleculeInfo' object is not subscriptable
```

I have resolved this.

christophlohrmann:
> > > I would like to handle all plotting in an alternative PR as it should be adjusted for many files not just the ones here but all Einstein and GK plots. However, for this PR, should the changes made to how the data is stored, i.e, msds not with the /6 be propagated to all Einstein computations? I think this is within the scope of the PR and should be done before merging. @PythonFZ @christophlohrmann
> > 
> > 
> > I have no problem with a partial rollout of the plotting, bit by bit we can adapt all calculators. As to the 6: If you know that the other MSDs are wrong, we should correct them as fast as possible
> 
> There are 5 Einstein calculators and there will likely be a /6 in all of them with exception now of course of the self-diffusion. This is probably also true for GK and a /3

since you have the new and improved fitting, it would make sense to use it in call calculators

SamTov:
> > > > I would like to handle all plotting in an alternative PR as it should be adjusted for many files not just the ones here but all Einstein and GK plots. However, for this PR, should the changes made to how the data is stored, i.e, msds not with the /6 be propagated to all Einstein computations? I think this is within the scope of the PR and should be done before merging. @PythonFZ @christophlohrmann
> > > 
> > > 
> > > I have no problem with a partial rollout of the plotting, bit by bit we can adapt all calculators. As to the 6: If you know that the other MSDs are wrong, we should correct them as fast as possible
> > 
> > 
> > There are 5 Einstein calculators and there will likely be a /6 in all of them with exception now of course of the self-diffusion. This is probably also true for GK and a /3
> 
> since you have the new and improved fitting, it would make sense to use it in call calculators

The new fitting is already in them all and the updated data storage. But not the 1/6 move or the plotting. Plotting is fine but I think the 1/6 should be updated before this is merged.

SamTov:
> > > > > I would like to handle all plotting in an alternative PR as it should be adjusted for many files not just the ones here but all Einstein and GK plots. However, for this PR, should the changes made to how the data is stored, i.e, msds not with the /6 be propagated to all Einstein computations? I think this is within the scope of the PR and should be done before merging. @PythonFZ @christophlohrmann
> > > > 
> > > > 
> > > > I have no problem with a partial rollout of the plotting, bit by bit we can adapt all calculators. As to the 6: If you know that the other MSDs are wrong, we should correct them as fast as possible
> > > 
> > > 
> > > There are 5 Einstein calculators and there will likely be a /6 in all of them with exception now of course of the self-diffusion. This is probably also true for GK and a /3
> > 
> > 
> > since you have the new and improved fitting, it would make sense to use it in call calculators
> 
> The new fitting is already in them all and the updated data storage. But not the 1/6 move or the plotting. Plotting is fine but I think the 1/6 should be updated before this is merged.

I have now done all of the Einstein calculators

SamTov:
@PythonFZ @christophlohrmann,  seems to an issue installing ZincHub, could you guys take a look?

PythonFZ:
> @PythonFZ @christophlohrmann, seems to an issue installing ZincHub, could you guys take a look?

Git or PyPi updated their package. Using `https` instead of `git` in 176d3ed fixed this issue. Unfortunately there still seems to be a memory issue with the notebooks.

SamTov:
> > @PythonFZ @christophlohrmann, seems to an issue installing ZincHub, could you guys take a look?
> 
> Git or PyPi updated their package. Using `https` instead of `git` in [176d3ed](https://github.com/zincware/MDSuite/commit/176d3ed9ccb87b1c606d0dc91912c7db2e9d7752) fixed this issue. Unfortunately there still seems to be a memory issue with the notebooks.

On Jupiter here in nbtest it is still a tag error but I can't understand why as in the branch it is correct. Did you get memory errors somewhere? All the big test run fine here.

PythonFZ:
> > > @PythonFZ @christophlohrmann, seems to an issue installing ZincHub, could you guys take a look?
> > 
> > 
> > Git or PyPi updated their package. Using `https` instead of `git` in [176d3ed](https://github.com/zincware/MDSuite/commit/176d3ed9ccb87b1c606d0dc91912c7db2e9d7752) fixed this issue. Unfortunately there still seems to be a memory issue with the notebooks.
> 
> On Jupiter here in nbtest it is still a tag error but I can't understand why as in the branch it is correct. Did you get memory errors somewhere? All the big test run fine here.

I was distracted by the memory warnings. But indeed it was a missing tag. Hopefully fixed in 3c3aed7

SamTov (approved):
Works well now.

PR #497, opened 2022-02-11
replace species dict with `SpeciesInfo` for #392

### LINKED ISSUE #392: experiment.species should be a dataclass

PythonFZ:
Fix #392 

- change `species` from `dotdict` to `SpeciesInfo`
- `experiment.species` remains a dictionary so it can be accessed via `species[key]: SpeciesInfo`. There is some redundant information here about the name being the key in the dict and also the `SpeciesInfo.name` but this simplifies the access and prohibits using `[x for x in SpeciesInfo if x.name == key]` all over the place. 
- add `groups` to SpeciesInfo
- replace the dict of `molecules` with `SpeciesInfo`

# TODOs
we should look into that part of the code again and decide what information we should store in SQL and what in HDF5 because there are some redudancies.


SamTov:
Can you elaborate on the redundancies @PythonFZ ?

PythonFZ:
> Can you elaborate on the redundancies @PythonFZ ?

Take the species for example. Currently we store information such as charge and mass in the SQL database but they might also exist in the H5MD database so we should decide where we want to store them.

SamTov:
> > Can you elaborate on the redundancies @PythonFZ ?
> 
> Take the species for example. Currently we store information such as charge and mass in the SQL database but they might also exist in the H5MD database so we should decide where we want to store them.

I agree we should consider alternatives. In principle though it is in hdf5 if it changes w.r.t time and SQL otherwise. In the case of charge it can be either and therefore can exist in both.

PythonFZ:
> > > Can you elaborate on the redundancies @PythonFZ ?
> > 
> > 
> > Take the species for example. Currently we store information such as charge and mass in the SQL database but they might also exist in the H5MD database so we should decide where we want to store them.
> 
> I agree we should consider alternatives. In principle though it is in hdf5 if it changes w.r.t time and SQL otherwise. In the case of charge it can be either and therefore can exist in both.

I do think that we should only store it in the HDF5 database. It can also store time-independent data.
But we should discuss this in more detail next week.

SamTov (approved):
Everything is good, there are some more places where this is broken though as evidenced in the tests. Once they pass it is ready.

christophlohrmann (approved):

mdsuite/database/simulation_database.py
```diff
@@ -97,6 +97,23 @@ def __eq__(self, other):
...
+    """Information about a Molecule
+
+    All the information of a species + groups
+
+    Attributes
+    ----------
+    groups: dict
+        A molecule specific dictionary for mapping the molecule to the
+        particles. TODO SamTov can you explain a bit more because I haven't found
+        a good docstring to copy/paste from map_molecules.py
```
PythonFZ, line 111: @SamTov Could you update the docstrings here?

PythonFZ: @SamTov could you update the docstrings here?

PythonFZ: + update notebook

SamTov: Resolved in [1aaa880](https://github.com/zincware/MDSuite/pull/497/commits/1aaa880c3a1bdeff4e715b2912d089de64877d33)

PR #500, opened 2022-02-15
compute volume from box array (not cached)

### LINKED ISSUE #499: Compute Volume dynamically and don't store it
Currently the experiment volume is computed in 
https://github.com/zincware/MDSuite/blob/68a475975e4c5d8b0ee9e0432ea06d2b55b3a1fc/mdsuite/experiment/experiment.py#L621
so if you have some `xyz` data and then add the cell via `experiment.cell = [13.3, 13.3, 13.3]` the volume is not updated.

I would suggest computing the volume on the fly via `volume = np.prod(cell)` and not storing it.
This would be even more imporant for simulations with changing box size.

PythonFZ:
Fix #499 

The volume is computed on the fly each time via `np.prod(box_array)`

We could add https://docs.python.org/3/library/functools.html#functools.cached_property but if the box_array changes this causes issues. Computing the product of 3 values is probably so fast, that any if/else cache is not required.

### DIFF
--- a/mdsuite/database/experiment_database.py
+++ b/mdsuite/database/experiment_database.py
@@ -83,7 +83,6 @@ class ExperimentDatabase:
     number_of_configurations = LazyProperty()
     number_of_atoms = LazyProperty()
     sample_rate = LazyProperty()
-    volume = LazyProperty()
     property_groups = LazyProperty()
 
     def __init__(self, project: Project, experiment_name):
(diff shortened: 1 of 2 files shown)


christophlohrmann (approved):
Good for now.

Should be looked over once we rework optional data handling (i.e. the approach here does not work if box_array is time-dependent and lives in the hdf database)

PythonFZ:
> Good for now.
> 
> Should be looked over once we rework optional data handling (i.e. the approach here does not work if box_array is time-dependent and lives in the hdf database)

I agree - I think this is something we can discuss (after) tomorrow.

SamTov (approved):
looks good.

PR #501, opened 2022-02-15
make time optional in extxyz

### LINKED ISSUE #498: Issue with reading `extxyz` files produced by ASE
**Describe the bug**
Can not add extxyz data to MDSuite

# Error message
```
~\AppData\Local\Temp/ipykernel_5328/2793412633.py in <module>
----> 1 project.add_experiment(
      2     name="LiCl2",
      3     timestep=0.001,
      4     temperature=900,
      5     units="metal",

c:\users\fabia\pycharmprojects\mdsuite\mdsuite\project\project.py in add_experiment(self, name, timestep, temperature, units, cluster_mode, active, simulation_data)
    232 
    233         if simulation_data is not None:
--> 234             self.experiments[name].add_data(simulation_data)
    235 
    236         return self.experiments[name]

c:\users\fabia\pycharmprojects\mdsuite\mdsuite\experiment\experiment.py in add_data(self, simulation_data, force, update_with_pubchempy)
    499         else:
    500             proc = _get_processor(simulation_data)
--> 501             self._add_data_from_file_processor(
    502                 proc, force=force, update_with_pubchempy=update_with_pubchempy
    503             )

c:\users\fabia\pycharmprojects\mdsuite\mdsuite\experiment\experiment.py in _add_data_from_file_processor(self, file_processor, force, update_with_pubchempy)
    535         database = Database(self.database_path / "database.hdf5")
    536 
--> 537         metadata = file_processor.metadata
    538         architecture = _species_list_to_architecture_dict(
    539             metadata.species_list, metadata.n_configurations

c:\users\fabia\pycharmprojects\mdsuite\mdsuite\file_io\file_read.py in metadata(self)
     62     def metadata(self):
     63         if self._metadata is None:
---> 64             self._metadata = self._get_metadata()
     65         return self._metadata
     66 

c:\users\fabia\pycharmprojects\mdsuite\mdsuite\file_io\extxyz_files.py in _get_metadata(self)
    134             )
    135             header_1 = file.readline()
--> 136             sample_rate = int(round(_get_time(header_1) - _get_time(header)))
    137 
    138         species_list = get_species_list_from_tabular_text_reader_data(

TypeError: unsupported operand type(s) for -: 'NoneType' and 'NoneType'
```

# Reproduce

```
project = mds.Project()
project.add_experiment(
    name="LiCl",
    timestep=0.001,
    temperature=900,
    units="metal",
    simulation_data="traj.extxyz"
)
```
with the following as content of `traj.extxyz`
[traj.txt](https://github.com/zincware/MDSuite/files/8067230/traj.txt)

### Header lines

- works: `Lattice="13.3 0.0 0.0 0.0 13.3 0.0 0.0 0.0 13.3" Properties=species:S:1:pos:R:3:forces:R:3 cutoff=-1.0 energy=-30664.29709663569 nneightol=0.0 i=0, pbc="T T T" time=0.000,`
- does not work: `Lattice="13.3 0.0 0.0 0.0 13.3 0.0 0.0 0.0 13.3" Properties=species:S:1:pos:R:3:momenta:R:3:forces:R:3 energy=-317.7220481319967 pbc="T T T"`

christophlohrmann:
Fix #498 

Bandaid fix

### DIFF
--- a/CI/unit_tests/project/test_project_add_experiment.py
+++ b/CI/unit_tests/project/test_project_add_experiment.py
@@ -25,15 +25,61 @@ Summary
 -------
 """
 import os
+import pathlib
 
 import MDAnalysis
 import numpy as np
 import pytest
+from ase import units
+from ase.calculators.emt import EMT
+from ase.io import write
+from ase.lattice.cubic import FaceCenteredCubic
+from ase.md.velocitydistribution import MaxwellBoltzmannDistribution
+from ase.md.verlet import VelocityVerlet
 from zinchub import DataHub
 
 import mdsuite as mds
 import mdsuite.file_io.chemfiles_read
 import mdsuite.file_io.lammps_flux_files
+
+
+@pytest.fixture(scope="session")
+def ase_md_extxyz(tmp_path_factory) -> pathlib.Path:
+    """
+    References
+    -----------
+    https://wiki.fysik.dtu.dk/ase/tutorials/md/md.html
+
+    Returns
+    -------
+    pathlib.Path
+        path to an ASE MD simulation
+
+    """
+    size = 3
+    atoms = FaceCenteredCubic(
+        directions=[[1, 0, 0], [0, 1, 0], [0, 0, 1]],
+        symbol="Cu",
+        size=(size, size, size),
+        pbc=True,
+    )
+    atoms.calc = EMT()
+    MaxwellBoltzmannDistribution(atoms, temperature_K=300)
+    dyn = VelocityVerlet(atoms, 5 * units.fs)
+
+    temporary_path = tmp_path_factory.getbasetemp()
+    filename = temporary_path / "traj.extxyz"
+
+    for i in range(20):
+        dyn.run(1)
+        write(
+            filename,
+            atoms,
+            format="extxyz",
+            append=True,
+        )
+
+    return filename
 
 
 @pytest.fixture(scope="session")
--- a/CI/unit_tests/project/test_project_add_experiment.py
+++ b/CI/unit_tests/project/test_project_add_experiment.py
@@ -156,6 +156,23 @@ def test_add_file_from_str(traj_files, tmp_path):
     assert list(project.experiments) == ["NaCl"]
 
 
+def test_add_file_from_str_extxyz(ase_md_extxyz, tmp_path):
+    """Check that adding extxyz files from an ASE md simulation works"""
+    os.chdir(tmp_path)
+    project = mds.Project()
+    project.add_experiment(
+        "NaCl",
+        simulation_data=ase_md_extxyz,
+        timestep=0.1,
+        temperature=300,
+    )
+
+    assert list(project.experiments) == ["NaCl"]
+    assert project.experiments["NaCl"].box_array == [10.83, 10.83, 10.83]
+    assert project.experiments["NaCl"].temperature == 300
+    assert list(project.experiments["NaCl"].species) == ["Cu"]
+
+
 def test_multiple_experiments(tmp_path):
     """Test the paths within the experiment classes
 
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -16,3 +16,4 @@ coveralls
 mdanalysis
 black
 znvis
+ase
--- a/mdsuite/file_io/extxyz_files.py
+++ b/mdsuite/file_io/extxyz_files.py
@@ -139,6 +139,10 @@ class EXTXYZFile(mdsuite.file_io.tabular_text_files.TabularTextFileProcessor):
             if time_0 is not None and time_1 is not None:
                 sample_rate = int(round(time_1 - time_0))
             else:
+                log.warning(
+                    "Could not read sample rate from file. Please adjust the sample rate"
+                    " manually if required."
+                )
                 sample_rate = None
 
         species_list = get_species_list_from_tabular_text_reader_data(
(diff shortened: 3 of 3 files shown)


PythonFZ:
Should we add a test? I can put the data from #498 into DataHub or we can generate it inside the test using ase.

christophlohrmann:
if generation of the data is not super slow, I would prefer moving away from tests that rely on remote data

PythonFZ:
> if generation of the data is not super slow, I would prefer moving away from tests that rely on remote data

I added a fixture in 4fffdc5 that we can use for the tests. It doesn't take any significant time.
I also had an issue reading some other files produced by ASE, so we could consider using https://github.com/zincware/MDSuite/blob/4fffdc57f6b9f41a87818e2a1bd7a61c3648852e/CI/unit_tests/project/test_project_add_experiment.py#L78 for testing some more file formats with a `@pytest.mark.parametrized` for example.

dev-requirements.txt
```diff
@@ -16,3 +16,4 @@ coveralls
 mdanalysis
 black
 znvis
+ase
```
SamTov, line 19: We could also remove a piece of information from the already existing files, I guess the MD is pretty fast but then we don't need such a big package in the dev-req

SamTov (commented):
Can we consider not running a simulation to get the data? Or are we going with no limits on what dev-req can contain. ASE is a huge package for something that could be done with only an edit of a file. We could even do that edit with a sed call as a pre-processing step on already existing data

dev-requirements.txt
PythonFZ: I would not worry about the dev-req size. We also have mdanalysis in there for example.
This way we also catch changes in ASE and we can use it to produce other files that we might want to be able to read.
ASE is about 4 MB compressed.

PythonFZ: What we could do, is put the generation of the files into DataHub / DataHub CI if you feel very uncomfortable with having it here. But this makes maintaining it more difficult.

SamTov: nope I am happy with that. Sorry I did not read the previous review comments before making my review so did not see that I asked opposing points.

SamTov (approved):
Happy with review comments. The only thing I would suggest is to make everything optional. ASE basically says that because there is no defined extxyz format everything is extxyz. For this reason, everything in the file should be wrapped in try excepts and the user told what is not present.

CI/unit_tests/project/test_project_add_experiment.py
```diff
@@ -25,17 +25,63 @@
...
+from ase.md.verlet import VelocityVerlet
 from zinchub import DataHub
 
 import mdsuite as mds
 import mdsuite.file_io.chemfiles_read
 import mdsuite.file_io.lammps_flux_files
 
 
+@pytest.fixture(scope="session")
+def ase_md_extxyz(tmp_path_factory) -> pathlib.Path:
```
PythonFZ, line 47: Need to add the actual test

PythonFZ (changes requested):
We need the test before we merge.

CI/unit_tests/project/test_project_add_experiment.py
christophlohrmann: who writes it?

PythonFZ: I did in [2509418](https://github.com/zincware/MDSuite/pull/501/commits/2509418d471c8659027b5616a36fc29d04897188)

PythonFZ (approved):

christophlohrmann:
I cannot officially approve but I approve

SamTov (changes requested):
Should we make everything a try except first?

If we will use the ase guidelines should we completely adopt them now?

PythonFZ:
> Should we make everything a try except first?
> 
> If we will use the ase guidelines should we completely adopt them now?

Can you provide an example file (from ASE) where the above mentioned code fails?

SamTov:
> > Should we make everything a try except first?
> > If we will use the ase guidelines should we completely adopt them now?
> 
> Can you provide an example file (from ASE) where the above mentioned code fails?

It is more about adopting the complete ASE standard, xyz files have existed much longer than ASE has. ASE states that as there is no rule for extxyz files any xy OR extxyz file will be considered extxyz. This means that if you have nothing in that comment line, one thing in that comment line, or some arbitrary thing in that comment line, it will be extxyz. So yes ASE should be able to print an extxyz file without full information if the data you gave it was incomplete. But furthermore, it is also valid for anyone using the format. Sometimes you will have viral and sometimes not. By your own words you want to adopt the ase standard and make it work, if this is true, there should be try and excepts over every property in the comment line.

mdsuite/file_io/extxyz_files.py
```diff
@@ -133,7 +133,13 @@ def _get_metadata(self):
...
             )
             header_1 = file.readline()
-            sample_rate = int(round(_get_time(header_1) - _get_time(header)))
+
+            time_0 = _get_time(header)
+            time_1 = _get_time(header_1)
+            if time_0 is not None and time_1 is not None:
+                sample_rate = int(round(time_1 - time_0))
+            else:
+                sample_rate = None
```
SamTov, line 146: Also, looking at it after a few days, there should be some comment here letting the user know that this has not been recorded as it is very important for any dynamic quantities being computed.

PythonFZ: [b8bb054](https://github.com/zincware/MDSuite/pull/501/commits/b8bb0541607989518264fb0d52aa4efe241c976c)

SamTov (approved):
I think all header lines have appropriate try and excepts with warnings but we should keep this in mind for future problems.

PR #502, opened 2022-02-22
Add a test for all notebooks in the examples directory

PythonFZ:
Add a CI run test for all notebooks in the examples directory


SamTov (approved):
great changes. As long as tests pass it is good to go.

PR #505, opened 2022-03-02
Update CI to use Python 3.10

PythonFZ:
Run tests on Python 3.10


SamTov:
Open3D does not yet support python 3.10 and so the dev-requirements are uninstallable. 

Monitor https://github.com/isl-org/Open3D/issues/4535

Essentially they are waiting for PyTorch to support 3.10 and then they will too.

PythonFZ:
@SamTov  having a `wontfix` badge here could give the false impression that we never want to support Python>=3.9.

SamTov:
With https://github.com/isl-org/Open3D/pull/5320 this should be fine soon. The wheels are out but they are waiting on the next stable release for the default. At this point we can trigger this PR again.

SamTov (approved):
Nice addition, as soon as the tests pass it is good to go.

Issue #510, opened 2022-03-10
tests fail on main


christophlohrmann:
Tests fail because DataHub has changed.

Possible solution: Make it such that you can use specific versions of DataHub and MDSuite says which version it needs.
With that you can change Datahub (and reflect the update with a version change in a PR on mdsuite) without affecting any other branch of mdsuite.

PythonFZ:
Will be fixed by introducing `tag` in https://github.com/zincware/ZincHub/pull/3

This will require the keyword `tag` on all occurrences of `zinchub`. It will also prohibit using data straight from main and instead require a new tag for all changes, ideally providing a description on why the data was updated.

PR #511, opened 2022-03-11
lammpstraj: catch empty header for traj with only one time step

### LINKED ISSUE #379: Handle single snapshot trajectories
For some computations we may want only a single snapshot eg. RDF, rings statistics (when implemented). 
Current reader gets the two first configurations to compute the sampling rate. 
Therefore, in the case of a single snapshot the code crashes. 

I have implemented it locally with a simple try/except block. 
However, I'd link this to the PR for the new file readers #372 since it makes more sense to do it in the new ones already.

christophlohrmann:
fix #379 for lammps
the other readers were able to handle single snapshots before.

~Test fail because of #510~


SamTov (approved):
thank you, looks good.

PR #515, opened 2022-04-11
fixed single snapshot feature for lammps files

### LINKED ISSUE #379: Handle single snapshot trajectories

Fratorhe:
Fix #379.
Current implementation gave a StopIteration exception when handling single snapshots. 

* Fixed using a try/except statement. 

I am developing a test case for graphs which use a single snapshot. 
It will be added soon, but I believe that it should be in another PR.

### DIFF
--- a/mdsuite/file_io/lammps_trajectory_files.py
+++ b/mdsuite/file_io/lammps_trajectory_files.py
@@ -234,15 +234,16 @@ class LAMMPSTrajectoryFile(mdsuite.file_io.tabular_text_files.TabularTextFilePro
             file, self.n_header_lines, start_at=0
         )
         time_step_0 = int(first_header[1])  # Time in first configuration
-        second_header = mdsuite.file_io.tabular_text_files.read_n_lines(
-            file, self.n_header_lines, start_at=self.n_header_lines + n_particles
-        )
         # catch single snapshot trajectory (second_header == [])
-        if not second_header:
+        try:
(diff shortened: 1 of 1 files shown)


christophlohrmann:
Looks good to me @PythonFZ do you know what is going on in CI?

PythonFZ:
> Looks good to me @PythonFZ do you know what is going on in CI?

Not really tbh

christophlohrmann:
maybe rerunning will solve it :)

SamTov:
> maybe rerunning will solve it :)

Had the same idea

SamTov:
```bash
urllib.error.URLError: <urlopen error [Errno 101] Network is unreachable>
```
This one issue but I don't think it is the same for all of them.

SamTov:
One error was this request thing, not too sure what it is or how it came up, the other was the batching, I guess the restructure of the calculators needs to happen so that the new database structure can be applied and batching redone as it seems that there are a lot of hard to reproduce bugs that plague it.

christophlohrmann (approved):

PR #522, opened 2022-05-17
automatic `__version__` from package version

PythonFZ:
Fix #368

*  Use `importlib` to determine package version
* add test (inspired by poetry) to ensure version changes are covered precisely by the test suite


CI/test_mdsuite.py
```diff
@@ -0,0 +1,5 @@
+from mdsuite import __version__
+
+
+def test_version():
```
SamTov: What is the protocol for automatics updates here? Or is this going to need to be added to a release pipeline of sorts.

SamTov (commented):
I like the idea, I have one comment to be addressed but as soon as that is done and the tests pass it can go through.

SamTov:
What's the progress here @PythonFZ ?

Fratorhe:
What's the difference with #368 ?

PythonFZ:
> What's the progress here @PythonFZ ?

I have no controll of the failing tests and think this can be merged?

PythonFZ:
> What's the difference with #368 ?

``py
__version__ = importlib.metadata.version("mdsuite")
``

I think the `versioneer.py` is not required here, because we can use the `importlib` library shipped with python.

SamTov:
> > What's the progress here @PythonFZ ?
> 
> I have no controll of the failing tests and think this can be merged?

Why are they failing? From memory it was a Python 3.7 thing and I thought in that case we would just remove 3.7 from the support list. A bunch of them look like server request failures... One to code coverage and one to pubchempy...

PythonFZ:
> > > What's the progress here @PythonFZ ?
> > 
> > 
> > I have no controll of the failing tests and think this can be merged?
> 
> Why are they failing? From memory it was a Python 3.7 thing and I thought in that case we would just remove 3.7 from the support list. A bunch of them look like server request failures... One to code coverage and one to pubchempy...

I don't know but either way I think this should be merged regardless or a new PR should be made to fix these issues and then rebase/merge main into this one to make the tests pass.

PythonFZ:
@SamTov I don't know what is going on here? The Python 3.7 issue should be fixed but there seems to be some non-deterministic issue with some of the tests?

SamTov:
> @SamTov I don't know what is going on here? The Python 3.7 issue should be fixed but there seems to be some non-deterministic issue with some of the tests?

So pubchempy seems to have failed... but there is also a 11th or so decimal difference in the log tests, I am comfortable having 3 dp accuracy so that one should be fine.

PythonFZ:
> 7715502

I think this should be in a dedicated PR (for the future) don't you think so? It is not related to what this PR is supposed to do?
Either way thanks for fixing it. This should allow this PR to be finally merged.

SamTov:
> > [7715502](https://github.com/zincware/MDSuite/commit/7715502fba815a61305f745267b6f3d72f9eb960)
> 
> I think this should be in a dedicated PR (for the future) don't you think so? It is not related to what this PR is supposed to do? Either way thanks for fixing it. This should allow this PR to be finally merged.

Sure but the PR is not passing tests whereas this test does pass on other branches, so by this argument I would also say that whatever changed here that led to the test failing should be fixed before allowing the PR to pass. We could do a separate PR that merges into this branch and then all into main, perhaps the `best practice` for the future?

In any case, dealing with pubchempy issue is a bigger thing. Perhaps we just ship MDSuite with all of the data and avoid API calls...

PythonFZ:
> Sure but the PR is not passing tests whereas this test does pass on other branches, so by this argument I would also say that whatever changed here that led to the test failing should be fixed before allowing the PR to pass.

The changes made in this PR are minimal and do in no way affect the tests. I agree with your reasoning but in this case I strongly believe we are just lucky if the tests pass on other branches.

> We could do a separate PR that merges into this branch and then all into main, perhaps the `best practice` for the future?

I made a PR https://github.com/zincware/MDSuite/pull/526 for best practice guidlines where we can summarize them.

SamTov:
> > Sure but the PR is not passing tests whereas this test does pass on other branches, so by this argument I would also say that whatever changed here that led to the test failing should be fixed before allowing the PR to pass.
> 
> The changes made in this PR are minimal and do in no way affect the tests. I agree with your reasoning but in this case I strongly believe we are just lucky if the tests pass on other branches.
> 
> > We could do a separate PR that merges into this branch and then all into main, perhaps the `best practice` for the future?
> 
> I made a PR #526 for best practice guidlines where we can summarize them.

The test has been around for a very long time and never failed... I would argue we are unlucky to see it fail here. Or something has gone wrong.

Nonetheless, if the tests pass then we can merge this.

PythonFZ:
> > > Sure but the PR is not passing tests whereas this test does pass on other branches, so by this argument I would also say that whatever changed here that led to the test failing should be fixed before allowing the PR to pass.
> > 
> > 
> > The changes made in this PR are minimal and do in no way affect the tests. I agree with your reasoning but in this case I strongly believe we are just lucky if the tests pass on other branches.
> > > We could do a separate PR that merges into this branch and then all into main, perhaps the `best practice` for the future?
> > 
> > 
> > I made a PR #526 for best practice guidlines where we can summarize them.
> 
> The test has been around for a very long time and never failed... I would argue we are unlucky to see it fail here. Or something has gone wrong.
> 
> Nonetheless, if the tests pass then we can merge this.

The tests also fail here https://github.com/zincware/MDSuite/pull/526 - maybe a package version changed  recently.

SamTov (approved):
looks great

PR #525, opened 2022-06-03
Fix GK data saving.

### LINKED ISSUE #524: Integration range parameter is not what get's stored in the SQL database
**Describe the bug**
In GK computations users can set an integration range to inform the calculator where to measure the value from. This value is displayed in the plot but not stored in the SQL database.

**To Reproduce**
Perform GK on diffusion and change integration range.

**Solution** 

Add it to the SQL storage.

SamTov:
# Fix #524 

In this PR I correct how data is added to the SQL database in the GK calculators. Previously, the final integration value was stored whilst the integration range value was printed. This is corrected and the enhanced visualization added to the GK calculators.

## Added tests:
None yet but before converted to real PR there will be some tests to ensure the data is saved correctly.

## TODO

* ~~Test IC~~
* Add tests in integration test suite.


CI/integration_tests/calculators/test_green_kubo_ionic_conductivity.py
```diff
@@ -65,4 +66,12 @@ def test_project(traj_file, true_values, tmp_path):
 
     computation = project.run.GreenKuboIonicConductivity(plot=False)
 
+    # Time is wrong in the test data
```
christophlohrmann, line 69: Can you add the correct time to the true values?
Is it possible to write this test without the need for external data?

CI/integration_tests/calculators/test_green_kubo_ionic_conductivity.py
```diff
@@ -65,4 +66,12 @@ def test_project(traj_file, true_values, tmp_path):
...
+    computation["NaCl"]["System"].pop("time")
+    computation["NaCl"]["System"].pop("integral")
+    computation["NaCl"]["System"].pop("integral_uncertainty")
+
+    true_values["System"].pop("time")
+    true_values["System"]["acf"] = (np.array(true_values["System"]["acf"]) / 500).tolist()
```
christophlohrmann, line 75: What's that 500?

CI/integration_tests/calculators/test_green_kubo_ionic_conductivity.py
SamTov: Yep! Before merging I'd like to remove external data here, the time is saved as essentially all zeros... I am looking for an analytical solution

CI/integration_tests/calculators/test_green_kubo_ionic_conductivity.py
SamTov: There was a multiply by data range and then what followed after the saving was a divide by data range so there was no effect, I got rid of them both but the data saved on ZnHub had the redundant prefacto which in this specific case is 500. I will remove the online data for the test and try to find a best analytical solution

christophlohrmann: Perfect.

CI/integration_tests/calculators/test_green_kubo_ionic_conductivity.py
SamTov: To be solved in with #529 associated PR

SamTov:
@christophlohrmann it is at the state we discussed today.

SamTov:
Had the pubchempy server request failure.

CI/integration_tests/calculators/test_green_kubo_ionic_conductivity.py
```diff
@@ -65,4 +66,12 @@ def test_project(traj_file, true_values, tmp_path):
 
     computation = project.run.GreenKuboIonicConductivity(plot=False)
 
+    # Time is wrong in the test data
+    computation["NaCl"]["System"].pop("time")
+    computation["NaCl"]["System"].pop("integral")
+    computation["NaCl"]["System"].pop("integral_uncertainty")
+
+    true_values["System"].pop("time")
```
PythonFZ, line 74: Should we make an issue for this, so we don't forget?

mdsuite/calculators/green_kubo_ionic_conductivity.py
```diff
@@ -169,6 +175,7 @@ def _calculate_prefactor(self):
         -------
 
         """
+        # TODO improve docstring
```
PythonFZ, line 178: TODO improve docstring

mdsuite/calculators/green_kubo_ionic_conductivity.py
```diff
@@ -230,39 +223,80 @@ def _post_operation_processes(self):
         -------
 
         """
-        result = self.prefactor * np.array(self.sigma)
-
+        self.sigmas = np.array(self.sigmas)
```
PythonFZ: It should work without this line I think. `np.mean` should convert it to a `np.ndarray`

mdsuite/calculators/green_kubo_ionic_conductivity.py
```diff
@@ -230,39 +223,80 @@ def _post_operation_processes(self):
...
+        sigma = np.mean(self.sigmas, axis=0)
+        sigma_SEM = np.std(self.sigmas, axis=0) / np.sqrt(len(self.sigmas))
+
+        self.acfs = np.array(self.acfs)
```
PythonFZ: same as above

mdsuite/calculators/green_kubo_ionic_conductivity.py
```diff
@@ -230,39 +223,80 @@ def _post_operation_processes(self):
...
+        acf = np.mean(self.acfs, axis=0)
+        ionic_conductivity = self.prefactor * sigma[self.args.integration_range - 1]
+        ionic_conductivity_SEM = (
+            self.prefactor * sigma_SEM[self.args.integration_range - 1]
+        )
         data = {
-            self.result_keys[0]: np.mean(result).tolist(),
-            self.result_keys[1]: (np.std(result) / np.sqrt(len(result))).tolist(),
+            self.result_keys[0]: [ionic_conductivity],
+            self.result_keys[1]: [ionic_conductivity_SEM],
```
PythonFZ, line 235: I guess in the long run it would make sense to allow for plain `np.ndarrys` to be saved in data which might not be possible at the moment. 
In this case is there a reason to use `[ionic_conductivity]` instead of `ionic_conductivity.tolist()` ?

mdsuite/calculators/green_kubo_ionic_conductivity.py
```diff
@@ -230,39 +223,80 @@ def _post_operation_processes(self):
...
-                * self.experiment.units["time"],
-                y_data=np.array(val[self.result_series_keys[1]]),
-                title=(
-                    f"{val[self.result_keys[0]]: 0.3E} +-"
-                    f" {val[self.result_keys[1]]: 0.3E}"
+            # Compute vacf line
+            fig.line(
+                time,
+                acf,
+                color="#003f5c",
```
PythonFZ, line 265: If we make this an MDSuite default color maybe have it as `mds.utils.colors.DARKBLUE` ?

mdsuite/calculators/green_kubo_ionic_conductivity.py
```diff
@@ -230,39 +223,80 @@ def _post_operation_processes(self):
...
+                legend_label=(
+                    f"{selected_species}: {val[self.result_keys[0]][0]: 0.3E} +-"
+                    f" {val[self.result_keys[1]][0]: 0.3E}"
```
PythonFZ, line 268: Should we add units here?

PythonFZ (changes requested):
Just a few smaller comments.

For the notebook it currently shows `Could not load dynamic library 'libcudnn.so.8'`. Should we run it on a machine with cuda installed so that these errors don't occur?
- It should be `data["NaCl"]["Na"]["diffusion_coefficient"]` not `data["NaCl"].data_dict["Na"]["diffusion_coefficient"]`
- maybe print `KCl_data = KCl_file.get_file('.')[0]` so that a user not using datahub knows what to pass the `simulation_data`

CI/integration_tests/calculators/test_green_kubo_ionic_conductivity.py
SamTov: So I added some of the new LJ data stuff which I will make  a PR for after the API change where we will have data generate OTF in the CI rather than stored somewhere

mdsuite/calculators/green_kubo_ionic_conductivity.py
SamTov: ionic conductivity here is just a float, if it is converted to a list with the call then no reason. In any case ,I am not sure why this is a list anyway.. it was in previous calculators so I left it. However, a lot of the saving stuff will be updated soon and I'd love to have this PR moved through as it is slowing down API changes from being completed.

mdsuite/calculators/green_kubo_ionic_conductivity.py
SamTov: It's a good idea, added in #531

mdsuite/calculators/green_kubo_ionic_conductivity.py
SamTov: Is it in the label? Might be a tight squeeze. I'd also like to redo the plotting soon too in the new changes so possibly a different PR

mdsuite/calculators/green_kubo_ionic_conductivity.py
SamTov: This will also be changed in the new API for which there is a current PR waiting on this PR

SamTov: This will also be changed in the new API for which there is a current PR waiting on this PR

SamTov:
I think the data dict stuff should be changed in all calculators and tests and areas so perhaps we also do this in a seperate PR that focuses on it.

mdsuite/calculators/green_kubo_ionic_conductivity.py
PythonFZ: I guess then it should just be `self.result_keys[0]: ionic_conductivity` because I also don't think it makes sense that it is a list.

PythonFZ (approved):

PR #532, opened 2022-07-01
Add correct use of Units dataclass

PythonFZ:
Currently the `Units` dataclass is not properly used. This changes that.


SamTov (commented):
A huge effort! Thanks for that it looks a lot cleaner and much better to work with. I would wait for tests to pass to officially accept it but it looks great

PythonFZ:
> A huge effort! Thanks for that it looks a lot cleaner and much better to work with. I would wait for tests to pass to officially accept it but it looks great

It was quite repetitive and easy to do. So the perfect task for me at the moment. The tests pass now because I adapted the test.

SamTov (approved):
Thanks for updating the tests.

PR #534, opened 2022-07-01
Add Colours dataclass

### LINKED ISSUE #531: Add a set of MDSuite colours
First mention by @PythonFZ, we should add some special set of MDSuite defined colours to that they are more standard and easier to user.

PythonFZ:
Fix #531 

* add a dataclass that contains all hex codes for the used colours inside mdsuite
* change all occuring hex codes to dataclass attributes


SamTov (approved):
I think this is a great addition, thanks for taking care of that.

PR #535, opened 2022-07-01
 create dependabot.yml

PythonFZ:
Include dependabots automatic PRs to update package versions.


SamTov (approved):
Should be a great addition, I like it!

PR #538, opened 2022-07-20
Restrict experiment names

### LINKED ISSUE #537: prohibit experiment names starting with a number
**Describe the bug**

If you use something like

```py
project.add_experiment(
        name="273K_ML",
        timestep=0.01,
        temperature=273.0,
        units="metal",
        simulation_data=file_reader
    )
```

you can not access it via

```py
project.experiments.273K_ML
```

There are two possible solutions:
- rename experiments automatically
- prohibit names starting with a number

I'm in favor of raising an error when someone tries to use an experiment name starting with a number.

PythonFZ:
Fix #537 

# Summary
- restrict experiment names
- rename `Experiment.experiment_name` to `Experiment.name`
- rename `Experiment.experiment_path` to `Experiment.path`
- remove `Experiment.logfile_path`

# Open questions

What is the idea behind having dedicated database_path / figures_path? Currently database_path is set equal to path and I don't know if there are any tests that allow for changing it.

### DIFF
--- a/CI/unit_tests/experiment/test_Experiment.py
+++ b/CI/unit_tests/experiment/test_Experiment.py
@@ -1,0 +1,8 @@
+import pytest
+
+from mdsuite.experiment.experiment import Experiment
+
+
+def test_experiment_name():
+    with pytest.raises(ValueError):
+        Experiment(experiment_name="250K", project=None)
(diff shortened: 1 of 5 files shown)


SamTov (approved):
Looks like nice changes to the code.

PR #507, opened 2022-03-07
[WIP] Sam tov docs updates

SamTov:
Fix up docs to correctly display modules.

* closes #424

PR #543, opened 2022-07-21
add `mds.units.<unit_system>`

### LINKED ISSUE #533: Add `mds.units.METAL`
We could add the following enums for user / developer friendlier code:

```py
md_project.add_experiment(
    name="KCl", timestep=0.002, units='metal', temperature=1200.0, simulation_data=KCl_data
)
# vs
md_project.add_experiment(
    name="KCl", timestep=0.002, units=mds.units.METAL, temperature=1200.0, simulation_data=KCl_data
)
```

They already exist but can not be used from outside
https://github.com/zincware/MDSuite/blob/ffb8ae91b1fef8495a2e1b69e80ba4ad82979272/mdsuite/utils/units.py#L95

PythonFZ:
Fix #533

The added dataclasses are now frozen. To generate a new one from an existing one use `units = dataclasses.replace(mdsuite.units.SI, length=0.000007654, time=0.0056789)`

# TODO
- [x] test that everything works
- [ ] ~~add to documentation~~


SamTov:
Before updating docs let's finish up #507 It should be fine once the CI build passes.

SamTov:
@PythonFZ errors here appear to be PR related.

PythonFZ:
@SamTov the test should be fixed. Where and How would you put the new units into the documentation?

... Why did my pre-commit hook not work...

SamTov:
> @SamTov the test should be fixed. Where and How would you put the new units into the documentation?
> 
> ... Why did my pre-commit hook not work...

Pre-commit hook didn't work in what way? Usually it is a version issue unless the pre-commit file was uninstalled somehow on your PC.

As for documentation, I've been thinking we beef up the user guide section. So right now it has just a notebooks sub heading. Why not add a Code components sub heading and then a page on units?

General question about the PR, does it change how users can parse their own custom units?

PythonFZ:
The `pre-commit` stuff was  something I could resolve and not related to the PR.

> As for documentation, I've been thinking we beef up the user guide section. So right now it has just a notebooks sub heading. Why not add a Code components sub heading and then a page on units?

I like the idea.

> General question about the PR, does it change how users can parse their own custom units?

Should not have changed and would be something like:

```py
custom_units = mds.Units(time=3.14, length=2.718, ...)
```

SamTov:
> The `pre-commit` stuff was something I could resolve and not related to the PR.
> 
> > As for documentation, I've been thinking we beef up the user guide section. So right now it has just a notebooks sub heading. Why not add a Code components sub heading and then a page on units?
> 
> I like the idea.
> 
> > General question about the PR, does it change how users can parse their own custom units?
> 
> Should not have changed and would be something like:
> 
> ```python
> custom_units = mds.Units(time=3.14, length=2.718, ...)
> ```

Awesome, in hindsight it is actually in the tests and they pass so no problem

CI/integration_tests/calculators/test_green_kubo_self_diffusion_coefficients.py
```diff
@@ -71,9 +72,9 @@ def test_calculator(tmp_path):
     os.chdir(tmp_path)
     project = mds.Project()
     # introduce nontrivial units to make sure all conversions are correct
-    units = mdsuite.utils.units.si
-    units.length = 0.000007654
```
SamTov, line 75: A general comment here, should this just be a custom unit declaration rather than changing SI units?

mdsuite/__init__.py
```diff
@@ -43,6 +44,7 @@
     Report.__name__,
     "config",
     Molecule.__name__,
+    "units",
```
SamTov, line 47: can it be `units.__name__`?

SamTov (approved):
Really nice updates to the code, made a couple of comments that would be nice to be looked at but not really critical so I will approve it so as not to require an additional review

CI/integration_tests/calculators/test_green_kubo_self_diffusion_coefficients.py
PythonFZ: I just modified the original test. I guess it could also be a new unit but using this method we can show how to "inherit" from an existing one if this might ever be necessary.

mdsuite/__init__.py
PythonFZ: In this case it can't. `units` is the module and not the class. I don't think modules come with a `__name__`. We could add it just like with `__version__` but I rather keep it like this.

SamTov: I thought it still worked actually but it is all good in any case

CI/integration_tests/calculators/test_green_kubo_self_diffusion_coefficients.py
SamTov: Yeah I saw it is from the old test, sure we can leave it

PR #545, opened 2022-07-26
Fix test failure

SamTov:
Recently there has been a problem in one of the tests related to memory safety. I have tried to adjust the test parameters to reduce the effect of this statistical issue.


SamTov:
If we merge this and then update the other branches the tests will stop failing. In general, when the memory management is looked at we should look at some better measurement ideas as this can lead to these fluctuations.

PythonFZ (approved):
If it works, it works!

PR #551, opened 2022-08-05
fix n_atoms update when adding new data

### LINKED ISSUE #546: Number of atoms not showing in lammpstraj read
**Describe the bug**
It seems that the number of atoms is not being read in for at least the lammps trajectory. It is possible that this only occurs on experiment loading but I do not know for sure yet.

**To Reproduce**
Run the comparing molten salts example notebook and look at the number of atoms attribute.

**Expected behavior**
Number of atoms to be read in correctly and loaded correctly after the fact.

** Notes **

The urgency here isn't so high as this values isn't really used by anything it seems... In any case, it should be a simple fix.

christophlohrmann:
Fix #546


christophlohrmann:
Black CI sieht irgendwie kaputt aus @PythonFZ

SamTov:
> Black CI sieht irgendwie kaputt aus @PythonFZ

Usually it’s just a version discrepancy, but it passes here, is it failing locally?

SamTov (approved):
Looks great, the added test should keep things in check too. I will try to resolve the test issues today.

PR #552, opened 2022-08-08
Move distinct coefficient to jax for large speedup

### LINKED ISSUE #462: ``test_green_kubo_distinct_diffusion_coefficients.py`` is slow
... at least on my machine
when I run the testsuite with ``pytest -n8``, the testsuite is done in 2 mins and then waits another 20min for the one slow test to finally finish

### LINKED ISSUE #463: Improve distinct coefficient speed
Both Einstein and GK distinct diffusion coefficient calculators need to be improved for performance.

#### Einstein

* Move to einsum
* Look for other optimization i,e, symmetry arguments and reducing summations 

#### GK

* As above for Einstein
* Move to tensorflow implementation of cross-correlation

### LINKED ISSUE #530: Look into Jax vmap and pmap for some calculators
For some calculators, specifically, green Kubo diffusion, including distinct, I have seen 100x speed improvements using the vmap operation in Jax. The pmap operation would also actually provide some ability to device-level parallelise calculator operations so that we could say, e.g., a calculator can parallelise over on-node devices, e.g. multiple gpu or cpu and gpu, then something like dask and handles multiple nodes distribution. 

Would solve e.g #462

SamTov:
Fix #463 
Fix #462 
Fix #530 

* Moved green kubo distinct coefficient to Jax vmaping and make it 100x faster at least.
* moved einstein distinct coefficients to jax

Run the GK distinct test


mdsuite/calculators/green_kubo_distinct_diffusion_coefficients.py
```diff
@@ -170,6 +171,112 @@ def __call__(
 
         self.combinations = list(itertools.combinations_with_replacement(self.species, 2))
 
+    def _correlate(self, ds_a: np.ndarray, ds_b: np.ndarray) -> np.ndarray:
```
christophlohrmann: These functions do not use ``self`` and seem very useful for other applications as well. For the future it would be a good idea to have them as free functions available to all calculators

mdsuite/calculators/einstein_distinct_diffusion_coefficients.py
```diff
@@ -170,35 +170,110 @@ def __call__(
             species=species,
             fit_range=fit_range,
         )
-        self.time = self._handle_tau_values()
+        self.time = self._handle_tau_values() * self.experiment.units.time
 
         self.msd_array = np.zeros(self.args.data_range)  # define empty msd array
 
-    def msd_operation(self, ensemble: tf.Tensor, square: bool = True):
+    def _msd_operation(self, ds_a, ds_b) -> np.ndarray:
```
christophlohrmann: Also this would be better as a free function

dev-requirements.txt
```diff
@@ -17,3 +17,4 @@ mdanalysis
 black
 znvis
 ase
+astroid
```
christophlohrmann, line 20: what is this for?

mdsuite/calculators/einstein_distinct_diffusion_coefficients.py
```diff
@@ -170,35 +170,110 @@ def __call__(
             species=species,
             fit_range=fit_range,
         )
-        self.time = self._handle_tau_values()
+        self.time = self._handle_tau_values() * self.experiment.units.time
```
christophlohrmann, line 173: Would be nice if _handle_tau_values() actually did all the handling

mdsuite/calculators/einstein_distinct_diffusion_coefficients.py
```diff
@@ -288,8 +355,8 @@ def _post_operation_processes(self, species: Union[str, tuple] = None):
             error = np.sqrt(np.diag(covariance))[0]
 
             data = {
-                self.result_keys[0]: -1 / 6 * fit_values[0],
-                self.result_keys[1]: 1 / 6 * error,
+                self.result_keys[0]: -1 / 2 * fit_values[0],
```
christophlohrmann, line 307: Why did this prefactor change?

mdsuite/calculators/green_kubo_distinct_diffusion_coefficients.py
SamTov: I fully agree. I was thinking of a Green-Kubo parent class or even working out the past templating for these methods. it is the fist time I have put them into MDSuite but they are also capable of being mapped in parallel over devices, e.g. GPUs. So when we look into mulit-gpu or multi-node analysis this will be a nice thing to have

mdsuite/calculators/einstein_distinct_diffusion_coefficients.py
SamTov: Yes, we could either make it a free function and use it for the self and distinct diffusion, however, and as you may have seen, in order to compute the distinct values properly I actually compute the self diffusion. It would be a better idea I think to merge the two calculators into one and have the distinct calculation as an option or something. What do you think? Also, for now or for later? I would be a fan of doing this during the calculator restructure as these calculators are currently desired by some users.

dev-requirements.txt
SamTov: I have no idea at all. It threw a docs build error for this reason.

mdsuite/calculators/einstein_distinct_diffusion_coefficients.py
SamTov: Yeah I saw that too when adding it, I can update this.

mdsuite/calculators/einstein_distinct_diffusion_coefficients.py
SamTov: During the mapping I average over the dimension. Therefore I removed the dimension term here. I don't like this too much and we can just sum and leave the 6 here, what do you think?

mdsuite/calculators/green_kubo_distinct_diffusion_coefficients.py
christophlohrmann: That is 3 already steps ahead of what I was proposing, which was just to move the function as it is into sth like a ``math.py`` oder ``calculator_utils.py`` file

mdsuite/calculators/einstein_distinct_diffusion_coefficients.py
christophlohrmann: later is ok. as long as there is not too much code duplication

mdsuite/calculators/einstein_distinct_diffusion_coefficients.py
christophlohrmann: actually I like it more with the 2 here. But for these equations, a quick comment with the corresponding formulae would be nice so one can see where the numbers come from

SamTov: Yep fully agree, I will add that

mdsuite/calculators/green_kubo_distinct_diffusion_coefficients.py
SamTov: I think beforehand I would like to think about how that wil look at the best way to handle it as it is similar in the Einstein case. I will try to tackle it tomorrow.

SamTov: I have moved them to the calculator helper methods directory and added standalone tests

PythonFZ (approved):
From my point everything looks nice. Good job @christophlohrmann and @SamTov

PR #553, opened 2022-08-10
set pubchempy defaults to false

### LINKED ISSUE #527: Remove pubchempy dependency
> In any case, dealing with pubchempy issue is a bigger thing. Perhaps we just ship MDSuite with all of the data and avoid API calls...

_Originally posted by @SamTov in https://github.com/zincware/MDSuite/pull/522#issuecomment-1169964853_

We can ship a table of the static information required by MDSuite without requiring pubchempy to make MDSuite work offline and indepent of API requests.

SamTov:
Fix #527 

I am, just setting the default to False in the code here so that it will not come up in the tests. This will at least improve dev speed for the time being until we can decide on next steps. It can still be turned on during data loading, however, the only thing loaded at the moment is mass. I would suggest we leave it like this until the db move is over and then think of how to have a more immersive experience in terms of populating particle properties.

* Set pubchempy call default to False


PythonFZ:
I don't know how to review the changes. I think we will come back to the PR when we found a solution for the tests.

SamTov:
> I don't know how to review the changes. I think we will come back to the PR when we found a solution for the tests.

Changes wise it's just setting pubchempy to false. Test fail due to masses not being present, They can be set in the tests I just haven't had the time to do it.

Perhaps an alternative is also to use the downloaded data so that the property update can still run but always locally.

examples/notebooks/NaCl_walkthrough.ipynb
```diff
@@ -99,7 +81,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "2022-07-27 15:40:42,240 - INFO: Creating new project NaCl_Example\n"
+      "2023-01-22 21:54:00,097 - INFO: Creating new project NaCl_Example\n"
```
christophlohrmann: restart and clear all output

christophlohrmann (changes requested):

christophlohrmann (approved):

PR #555, opened 2022-08-22
Update setup.py

SamTov:
Updated setup.py for new release


PythonFZ (approved):

PythonFZ:
What do you think about adding a Changelog so we have a better overview of what has been updated?

PR #563, opened 2022-11-29
Sam tov memory scaling function

SamTov:
Added memory fraction as a configuration parameter

* Added  used memory fraction as a configuration parameter
* Updated memory manager to use this

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -46,7 +46,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.x'
+          python-version: '3.9'
       - name: Install package
         run: |
           pip install .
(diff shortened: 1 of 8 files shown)


.github/workflows/lint.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.x'
+          python-version: '3.9'
```
PythonFZ, line 49: Should be fixed on all linting?

mdsuite/utils/config.py
```diff
@@ -36,11 +36,15 @@ class Config:
     bokeh_sizing_mode: str
         The way bokeh scales plots.
         see bokeh / sizing_mode for more information
+    jupyter : bool
+            If true, jupyter is being used.
```
PythonFZ, line 40: What does this mean?

SamTov: If true, the environment is jupyter.

.github/workflows/lint.yaml
SamTov: Perhaps, we should probably open an issue / PR and discuss it as sometimes it is nice for it to be latest but a fixed one for linting is probably okay.

PythonFZ (approved):

PR #564, opened 2022-11-29
Sam tov fix memory scaling tests

SamTov:
Using #563 I am trying to fix GitHub CI jobs that seem to fail in minibatch limits

* Found that the latest TensorFlow release doesn't like using concat on scalars so this has been changed to stack.
* The Einstein averaging for remainder case was also wrong but this is fixed now and local testing demonstrated that it works.
* I could test the atomwise batching and it worked great! 

We should consider how best to add low memory cases. We can run all of the tests in a memory_fraction=0.9 and then a memory_fraction=0.01 and that should test all cases. @christophlohrmann @PythonFZ what do you think? It should probably come in a different PR once we decide on the best format for duplicating tests only with one configuration parameter changing

### DIFF
--- a/CI/integration_tests/calculators/test_radial_distribution_function.py
+++ b/CI/integration_tests/calculators/test_radial_distribution_function.py
@@ -32,7 +32,6 @@ from zinchub import DataHub
 import mdsuite
 import mdsuite as mds
 from mdsuite.utils.testing import assertDeepAlmostEqual
-mdsuite.config.memory_fraction = 1e-100
 
 
 @pytest.fixture(scope="session")
--- a/CI/integration_tests/calculators/test_radial_distribution_function.py
+++ b/CI/integration_tests/calculators/test_radial_distribution_function.py
@@ -29,7 +29,6 @@ import os
 import pytest
 from zinchub import DataHub
 
-import mdsuite
 import mdsuite as mds
 from mdsuite.utils.testing import assertDeepAlmostEqual
 
--- a/CI/integration_tests/calculators/test_radial_distribution_function.py
+++ b/CI/integration_tests/calculators/test_radial_distribution_function.py
@@ -29,8 +29,10 @@ import os
 import pytest
 from zinchub import DataHub
 
+import mdsuite
 import mdsuite as mds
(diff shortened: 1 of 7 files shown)


CI/integration_tests/calculators/test_radial_distribution_function.py
```diff
@@ -29,6 +29,7 @@
 import pytest
 from zinchub import DataHub
 
+import mdsuite
 import mdsuite as mds
```
PythonFZ: Should bei either one. I prefer plain `import mdsuite as mds`

mdsuite/calculators/einstein_diffusion_coefficients.py
```diff
@@ -184,17 +184,22 @@ def ensemble_operation(self, ensemble):
         msd = tf.math.squared_difference(
             tf.gather(ensemble, self.args.tau_values, axis=1), ensemble[:, None, 0]
         )
+        self.count += msd.shape[0]
         # average over particles, sum over dimensions
-        msd = tf.reduce_sum(tf.reduce_mean(msd, axis=0), axis=-1)
+        # msd = tf.reduce_sum(tf.reduce_mean(msd, axis=0), axis=-1)
+        msd = tf.reduce_sum(tf.reduce_sum(msd, axis=0), axis=-1)
```
PythonFZ, line 190: I don't know the shape here, but I assume something like `(a, b, c)` so it would be a little faster to use
`tf.reduce_sum(msd, axis=(0, -1))` or `tf.einsum("abc -> b", data)`

PythonFZ (approved):
Some suggestions but looks good.

mdsuite/calculators/einstein_diffusion_coefficients.py
SamTov: I think in the calculator restructure these have all been re-written so to avoid (more than will already exist) conflicts I would leave it for the time being.

CI/integration_tests/calculators/test_radial_distribution_function.py
SamTov: Yep, that should be changed

PR #565, opened 2022-11-30
Low memory tests

SamTov:
* Add tests for all calculators at 1 MB memory limit
* Added helper method to automatically compute memory fraction based on currently available machine memory
* Fixed failing tests in low memory limit


CI/integration_tests/calculators/test_angular_distribution_function.py
```diff
@@ -69,3 +70,21 @@ def test_project(traj_file, true_values, tmp_path):
...
+    os.chdir(tmp_path)
+    project = mds.Project()
+    project.add_experiment(
+        "NaCl", simulation_data=traj_file, timestep=0.002, temperature=1400
+    )
+
+    computation = project.run.AngularDistributionFunction(plot=False)
+
+    for item in computation["NaCl"].data_dict:
+        computation["NaCl"].data_dict[item].pop("max_peak")
```
PythonFZ: We should not use `.data_dict` anymore.
Why do you remove the `max_peak`?

CI/integration_tests/calculators/test_angular_distribution_function.py
```diff
@@ -69,3 +70,21 @@ def test_project(traj_file, true_values, tmp_path):
...
+
+    assertDeepAlmostEqual(computation["NaCl"].data_dict, true_values, decimal=-2)
+    mds.config.memory_fraction = 0.5
```
PythonFZ: I don't like the code duplication of changing the memory fraction back. Could we use a helpfer function for all tests like:

```python
@contextlib.contextmanager
def reduce_memory_fraction(memory_fraction: float = 0.1):
  default = mds.config.memory_fraction
  mds.config.memory_fraction = compute_memory_fraction(memory_fraction)

try:
  yield
finally:
  mds.config.memory_fraction = default
```

PythonFZ: This also reduced code duplication because we can write:

```python
@pytest.mark.parametrize("memory_fraction": (mds.config.memotry_fraction, 0.1))
def my_test(memory_fraction):
   with reduce_memory_fraction(memory_fraction):
      ...
```

mdsuite/calculators/green_kubo_self_diffusion_coefficients.py
```diff
@@ -196,11 +195,15 @@ def ensemble_operation(self, ensemble):
...
         )
+        self.count += vacf.shape[0]
 
         # average particles, sum dimensions
-        vacf = tf.reduce_sum(tf.reduce_mean(vacf, axis=0), -1)
-        self.sigmas.append(cumtrapz(vacf, x=self.time))
-        self.vacfs.append(vacf)
+        return_vacf = tf.reduce_sum(tf.reduce_sum(vacf, axis=0), -1)
+        self.sigmas.append(
+            cumtrapz(tf.reduce_sum(tf.reduce_mean(vacf, axis=0), -1), x=self.time)
```
PythonFZ, line 201: I don't understand, why you duplicated the code here? Also consider using `tf.reduce_sum(vacf, axis=(0, -1))` for performance reasons.

mdsuite/utils/helpers.py
```diff
@@ -23,3 +26,28 @@ class NoneType:
...
+
+    Returns
+    -------
+    memory fraction : float
+            What fraction of the current systems memory this value corresponds to.
+            If this number is above 1, it is clipped to 1
+    """
+    if total_memory is None:
+        total_memory = get_machine_properties()["memory"] / (1024.0**3)
+
```
PythonFZ, line 53: Just curious, does it really make a difference if you use `1024.0` instead of `1024`?

PythonFZ (changes requested):
I think a lot of code duplication can be avoided by using a context manager to  set the memory_fraction.

CI/unit_tests/utils/test_helpers.py
```diff
@@ -0,0 +1,55 @@
...
+        """
+        Test the compute memory fraction helper function.
+        """
+        # Assert for real fraction.
+        memory_fraction = compute_memory_fraction(desired_memory=1.0, total_memory=2.0)
+        assert memory_fraction == 0.5
+
+        # Assert for impossible fraction
+        memory_fraction = compute_memory_fraction(desired_memory=2.5, total_memory=2.0)
+        assert memory_fraction == 0.9
```
christophlohrmann: why is this the expexted behaviour?

christophlohrmann (changes requested):
I agree with @PythonFZ that the tests should not be duplicated. Context managers are nice (albeit a bit complicated), something like
```python
def test_stuff():
    do_actual_test()
    mds.set_memory(1mb)
    do_actual_test()
```
probably also does the trick

PythonFZ:
> I agree with @PythonFZ that the tests should not be duplicated. Context managers are nice (albeit a bit complicated), something like
> 
> ```python
> def test_stuff():
>     do_actual_test()
>     mds.set_memory(1mb)
>     do_actual_test()
> ```
> 
> probably also does the trick

The reason for the context manager was, that with `mds.set_memory(1mb)` we change a global MDSuite setting, so even if the test fails we need to make sure that we revert that change. I don't know if we can be certain that this happens if we don't use a context manager. This would not affect the CI, because here all tests have to pass  but for local testing it might cause unexpected behaviour when running tests with `pytest -n auto` and such a test fails.

CI/integration_tests/calculators/test_angular_distribution_function.py
SamTov: This was a long time ago. Most of the data we have saved is wrong and we need proper test data to check it. In the water data we kind of have this but as of yet, there is no nice thing we can use here. When these calculators were fixed a few months ago this was changed. I think removing data dict is in a different issue somewhere, I did not add this manually.

CI/integration_tests/calculators/test_angular_distribution_function.py
SamTov: This looks awesome, I would much prefer it be like this, could you do some of these so I see what you mean as I don't fully see how to implement that and have it run through all the tests but something like this is what I hoped I could do but I went for the simple way so I could actually test the calculators which uncovered some faults.

mdsuite/calculators/green_kubo_self_diffusion_coefficients.py
SamTov: One provides an uncertainty and one provides the actual acf which is normalized later to avoid false normalization when we have mini-batched or remainder ensembles. I could remove the sigmas line but then we would have no uncertainty on the prediction so thus far this is the best way.

mdsuite/utils/helpers.py
SamTov: I just prefer floats as they play better with the tensor libraries but no.

CI/unit_tests/utils/test_helpers.py
SamTov: In the function I write that if the value you requested is more than the maximum memory of the machine it sets the fraction to 0.9. This just tests that it identifies this and sets the memory fraction.

CI/integration_tests/calculators/test_angular_distribution_function.py
PythonFZ: You should be able to replace it with `computation["NaCl"][item].pop("max_peak")`

SamTov:
> > I agree with @PythonFZ that the tests should not be duplicated. Context managers are nice (albeit a bit complicated), something like
> > ```python
> > def test_stuff():
> >     do_actual_test()
> >     mds.set_memory(1mb)
> >     do_actual_test()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > probably also does the trick
> 
> The reason for the context manager was, that with `mds.set_memory(1mb)` we change a global MDSuite setting, so even if the test fails we need to make sure that we revert that change. I don't know if we can be certain that this happens if we don't use a context manager. This would not affect the CI, because here all tests have to pass but for local testing it might cause unexpected behaviour when running tests with `pytest -n auto` and such a test fails.

I like the idea but if you could write them or at least some it would be helpful as I am not sure of the best way to do it. Also, if you do that, see if you can replicate the CI error locally. It says that time isn't in the data that is loaded but this is not true...

CI/integration_tests/calculators/test_angular_distribution_function.py
PythonFZ: I added the context manager in [84a4c4b](https://github.com/zincware/MDSuite/pull/565/commits/84a4c4bebc2993f40741192d8b212ec764aad169) and applied it to the ADF test [51db65f](https://github.com/zincware/MDSuite/pull/565/commits/51db65ff9c3a583bf74bd91f643a0dab74e6f9a1)

SamTov:
@PythonFZ Thanks for moving all the calculators that's awesome. Looks a lot nicer now. I realise the issue is that the data is loaded and then popped once and then we try it again. I will fix it now.

CI/integration_tests/calculators/test_angular_distribution_function.py
SamTov: That's awesome thanks!

CI/integration_tests/calculators/test_angular_distribution_function.py
SamTov: I think we should move it all over at one time but I will be trying to add proper data for several tests next week and I will then re-write many of these.

mdsuite/calculators/green_kubo_ionic_conductivity.py
```diff
@@ -293,21 +283,16 @@ def plot_data(self, data):
...
+        """Run analysis."""
         self.check_input()
         # Compute the pre-factor early.
         self._calculate_prefactor()
 
         dict_ref = str.encode(
             "/".join([self.loaded_property.name, self.loaded_property.name])
         )
-
+        self.count = 0
```
PythonFZ, line 294: Nothing that needs to be changed, but making a run variable a class attribute somehow doesn't feel right.

PythonFZ (approved):

mdsuite/calculators/green_kubo_ionic_conductivity.py
SamTov: It is updated in methods throughout the class. An edit to that, at least for anatomist cases, in this case maybe not, but it has the same style as the others. It will all be removed soon. But in general yes i agree, that was just lazy copying.

christophlohrmann (approved):

PR #569, opened 2022-12-05
Change compression algorithm

SamTov:
Found a case where the compression algorithm is too punishing. For scientific notation it is very easy to get values on a scale < 1e-6. Therefore I propose changing back to gzip where no data can be lost.


christophlohrmann:
do we not use h5 and sql to store data? Where is that compression applied?

PythonFZ:
Does this break backwards compatibility?

SamTov:
> do we not use h5 and sql to store data? Where is that compression applied?

Compression is applied to the hdf5 datasets

SamTov:
> Does this break backwards compatibility?
It shouldn't no.

christophlohrmann:
> > do we not use h5 and sql to store data? Where is that compression applied?
> 
> Compression is applied to the hdf5 datasets

I thought hdf5 was a binary format.
Anyhow I think we should always go for lossless compression in any data storage

SamTov:
> > > do we not use h5 and sql to store data? Where is that compression applied?
> > 
> > 
> > Compression is applied to the hdf5 datasets
> 
> I thought hdf5 was a binary format. Anyhow I think we should always go for lossless compression in any data storage

It is but you can still apply compression to it. I agree and have found some cases that really require it.

PythonFZ (approved):
Is it possible to test this? Otherwise I guess the tests fail because of something else?

SamTov:
> Is it possible to test this? Otherwise I guess the tests fail because of something else?

So the tests failing in the IC could actually be because of the compression change but I need to check what is happening. In general it does not break backwards compatibility though because the compression algorithm is checked by h5py before loading the data. Otherwise one would need to pass the algorithm to the data loader.

I will try to resolve these tests failures locally.

PythonFZ (approved):

PR #577, opened 2023-01-23
Update pytest.yml

PythonFZ:
Remove CI Tests on PR.


SamTov (approved):
I like the changes and we can add the ruff constrains back at a later stage.

@PythonFZ looks like we will need to make the other PRs merge into this one to have it done properly. That will include the numpy type changes but that is probably all.

PR #581, opened 2023-01-27
Move system transformations to new folder

### LINKED ISSUE #568: System wide observables and data loading
**Describe the bug**
There is a mismatch in naming convention for system wide observables depending on whether they have been loaded from a file or produced using a transformation.

In the case that it is loaded from a file they are placed under the `Observables` group. However, if one produces something system-wide like an ionic current it will be placed directly, into an `Ionic_Current` group. This means that if someone wants to use an observable that was read in from a file for a calculation, it will fail as it does not know from where it should load the data.

**Expected behaviour**
The convention should be clear and in the move to H5MD structure, should likely be the `Observables` group convention. However, none of the calculators use this convention...

SamTov:
Fix #568 

I have made it such that all transformations put data into a group called "Observables" rather than one based on their own name. This is in line with the file readers when they load in a flux file. It will allow us to avoid have different loading places for the same observable.

At the moment there is no specific test for this, the fact that the transformation tests pass show that the paths have been correctly re-routed but we should add a test for a flux file and a transformation putting data into the same group I guess. This will be handled in #571


PythonFZ (approved):
Looks good, tests are passing.

PR #582, opened 2023-01-31
Update .pre-commit-config.yaml

PythonFZ:
- fix ruff version
- fix isort version
- fix black version


SamTov:
This will fix the inconsistent ruff calls?

PythonFZ:
> This will fix the inconsistent ruff calls?

it should fix all errors related to e.g. `D401` not found. I haven't encountered them in the MDSuite project so it is difficult to tell if this is really the cause but  it could be.

SamTov (approved):
Awesome thank you!

PR #583, opened 2023-01-31
Python fz add dataset keys

PythonFZ:
Add the following to avoid using the string `"Observables"` everywhere.
```python
class DatasetKeys(metaclass=_FrozenCls):
    """Class to hold the keys for the datasets."""

    OBSERVABLES = "Observables"
```


SamTov (approved):
Fantastic! Thanks!

PR #584, opened 2023-02-01
Add `zincware` badge

PythonFZ:
Add bade from https://shields.io/ [zincware](https://github.com/zincware)


SamTov (approved):
It's exactly what we have been missing

PR #586, opened 2023-02-10
remove plot rescaling to fix bug

### LINKED ISSUE #585: Plots are not displayed properly in jupyter
**Describe the bug**
There seems to be a change in the latest bokeh plots version that causes plots to be not displayed property.

Should we try to fix the plots in Bokeh or move on to plotly?

SamTov:
Fix #585 

Removed manual scaling of the jupyter plots.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.x'
+          python-version: '3.10'
       - name: Install isort
         run: |
           pip install isort==5.12.0
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -15,7 +15,6 @@ jobs:
       fail-fast: false
       matrix:
         python-version: 
-        - 3.7
         - 3.8
         - 3.9
         - "3.10"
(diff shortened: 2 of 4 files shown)


PythonFZ (commented):
> So with Bokeh 3.03 and BokehJS 3.03 it produces these mini plots.

Should we fix the Bokeh version then to >= 3.03 ?

SamTov:
> > So with Bokeh 3.03 and BokehJS 3.03 it produces these mini plots.
> 
> Should we fix the Bokeh version then to >= 3.03 ?

I thought about it, I don't remember what the exact issue was for the rescaling. I think the plots were just larger in jupyter environments but they weren't broken. But I don't have any strong opinion it isn't a breaking change to restrict it or not.

PythonFZ:
> > > So with Bokeh 3.03 and BokehJS 3.03 it produces these mini plots.
> > 
> > 
> > Should we fix the Bokeh version then to >= 3.03 ?
> 
> I thought about it, I don't remember what the exact issue was for the rescaling. I think the plots were just larger in jupyter environments but they weren't broken. But I don't have any strong opinion it isn't a breaking change to restrict it or not.

I think we should fix it. It goes to a newer major release of Bokeh and if we fix it, we are on the safe side. I'll make the change and then we can merge.

PythonFZ (approved):

PR #587, opened 2023-02-10
Allow for new groups to be added after DS creation.

### LINKED ISSUE #570: Adding new data in different group impossible after first run
**Describe the bug**
Currently if I run the following the code will fail:

```python
headers = {
     mdsuite_properties.thermal_flux.name : [
        "Jx",
        "Jy",
        "Jz",
    ]
}
reader = LAMMPSFluxFile(
    file_path="log.lammps",
    sample_rate=1, 
    box_l=3 * [28.9], 
    custom_data_map=headers, 
    n_header_lines=1
)
exp = project.add_experiment(
    name="lj", 
    timestep=1.0, 
    units=mds.units.METAL, 
    simulation_data=["gk_data.lammpstraj", reader],
    temperature=118
)
```

So here in the experiment creation I am adding some trajectory data and then after the experiment is built, I try to add some flux data. The issue is in the experiment code. When an `add_data` call is made, the only thing the experiment does is check if the database exists and if so, simply tries to expand the already built datasets which leads to an error if the dataset and group does not exist.

SamTov:
Fix #570 

* Check if a group in the new data is different and add it to the database if so
* Added a test to check that it is working

## TODO

- [x] Add starting configuration check

~~The next problem comes in actually adding the data.~~

```python
for i, batch in enumerate(file_processor.get_configurations_generator()):
    database.add_data(chunk=batch, start_idx=self.number_of_configurations)
    self.number_of_configurations += batch.chunk_size
```
~~This code takes the first location in the database to be the current number of configurations. For some groups this will be different now. @christophlohrmann Some ideas for approaching this would be awesome.~~

I will try to store the last index as a dataset attribute making the need for passing the starting index redundant. It would also give us a clever way to over-write data.

This is resolved now and the PR is ready so long as tests are passing

### DIFF
--- a/mdsuite/database/simulation_database.py
+++ b/mdsuite/database/simulation_database.py
@@ -328,7 +328,7 @@ class Database:
 
         return architecture
 
-    def add_data(self, chunk: TrajectoryChunkData, start_idx: int):
+    def add_data(self, chunk: TrajectoryChunkData):
         """
         Add new data to the dataset.
 
@@ -344,14 +344,15 @@ class Database:
         chunk_data = chunk.get_data()
 
         with hf.File(self.path, "r+") as database:
-            stop_index = start_idx + chunk.chunk_size
-
             for sp_info in chunk.species_list:
                 for prop_info in sp_info.properties:
                     dataset_name = f"{sp_info.name}/{prop_info.name}"
                     write_data = chunk_data[sp_info.name][prop_info.name]
 
                     dataset_shape = database[dataset_name].shape
+                    start_idx = database[dataset_name].attrs["starting_index"]
+                    stop_index = start_idx + chunk.chunk_size
+
                     if len(dataset_shape) == 2:
                         # only one particle
                         database[dataset_name][start_idx:stop_index, :] = write_data[
@@ -370,6 +371,7 @@ class Database:
                             "dataset shape must be either (n_part,n_config,n_dim) or"
                             " (n_config, n_dim)"
                         )
+                    database[dataset_name].attrs["starting_index"] += chunk.chunk_size
 
     def resize_datasets(self, structure: dict):
         """
@@ -490,6 +492,8 @@ class Database:
                     compression="gzip",
                     chunks=True,
                 )
+                dataset = database[dataset_path]
+                dataset.attrs["starting_index"] = 0
 
     def _add_group_structure(self, structure: dict):
         """
--- a/mdsuite/database/simulation_database.py
+++ b/mdsuite/database/simulation_database.py
@@ -350,22 +350,24 @@ class Database:
                     write_data = chunk_data[sp_info.name][prop_info.name]
 
                     dataset_shape = database[dataset_name].shape
-                    start_idx = database[dataset_name].attrs["starting_index"]
-                    stop_index = start_idx + chunk.chunk_size
+                    start_index = database[dataset_name].attrs["starting_index"]
+                    stop_index = start_index + chunk.chunk_size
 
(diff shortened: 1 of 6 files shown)


mdsuite/database/simulation_database.py
```diff
@@ -398,14 +400,19 @@ def resize_datasets(self, structure: dict):
...
-                else:
-                    axis = 1
-                    expansion = dataset_information[1] + db[identifier].shape[1]
+                try:
+                    if len(dataset_information[:-1]) == 1:
+                        axis = 0
+                        expansion = dataset_information[0] + db[identifier].shape[0]
+                    else:
+                        axis = 1
+                        expansion = dataset_information[1] + db[identifier].shape[1]
```
PythonFZ: `expansion = dataset_information[axis] + db[identifier].shape[axis]`

mdsuite/database/simulation_database.py
```diff
@@ -344,14 +344,15 @@ def add_data(self, chunk: TrajectoryChunkData, start_idx: int):
...
         with hf.File(self.path, "r+") as database:
-            stop_index = start_idx + chunk.chunk_size
-
             for sp_info in chunk.species_list:
                 for prop_info in sp_info.properties:
                     dataset_name = f"{sp_info.name}/{prop_info.name}"
                     write_data = chunk_data[sp_info.name][prop_info.name]
 
                     dataset_shape = database[dataset_name].shape
+                    start_idx = database[dataset_name].attrs["starting_index"]
```
PythonFZ: `start_idx / stop_index`: Just a comment, but I think either `idx` or `index` should be used. I'm in favour of `index`

PythonFZ (changes requested):
only small changes. I'll review again as soon as the tests pass.

mdsuite/database/simulation_database.py
SamTov: Great catch, I have updated it in https://github.com/zincware/MDSuite/pull/587/commits/34fc9fa9394ab1836c2b71e755d76a635cf2b3f3

mdsuite/database/simulation_database.py
SamTov: Yep, I agree, fixed in https://github.com/zincware/MDSuite/pull/587/commits/34fc9fa9394ab1836c2b71e755d76a635cf2b3f3

mdsuite/database/simulation_database.py
```diff
@@ -398,14 +402,19 @@ def resize_datasets(self, structure: dict):
...
-                if len(dataset_information[:-1]) == 1:
-                    axis = 0
-                    expansion = dataset_information[0] + db[identifier].shape[0]
-                else:
-                    axis = 1
-                    expansion = dataset_information[1] + db[identifier].shape[1]
+                try:
+                    if len(dataset_information[:-1]) == 1:
+                        axis = 0
+                        expansion = dataset_information[axis] + db[identifier].shape[axis]
```
PythonFZ: Can you move this outside the `if .. else` ?

PythonFZ (approved):
Looks good.
Please see my comment before merging.

mdsuite/database/simulation_database.py
SamTov: Which part?

SamTov: resolve [move statement out of if else](https://github.com/zincware/MDSuite/pull/587/commits/bfbf4146cc8ea2477f68a8f800f911e0bf1540f2)

PR #588, opened 2023-02-13
update readme with disclaimer

SamTov:
Just added a disclaimer to the readme.


PythonFZ (approved):
I hope it will never be important to us.

PR #589, opened 2023-02-14
add DOI badge

PythonFZ:
add [DOI](https://doi.org/10.1186/s13321-023-00687-y)

alternatives:
- [DOI](https://doi.org/10.1186/s13321-023-00687-y)
- [DOI](https://doi.org/10.1186/s13321-023-00687-y)


SamTov (approved):
Thanks for that! I like the first style DOI

SamTov:
We will also need to update the headers of the files, but let's leave that for now.

PR #597, opened 2023-05-16
use pre-commit.ci

PythonFZ:
Instead of using Github action runners for linting and a combination off `requirements.txt`, `pre-commit-config.yaml` and CI code we combine everything at a single place and use https://pre-commit.ci

It will also make PR to update the tools.


dev-requirements.txt
```diff
@@ -10,12 +10,9 @@ sphinx-copybutton
 IPython
 nbsphinx-link
 git+https://github.com/zincware/ZincHub.git
-isort
 coverage
 coveralls
 mdanalysis
-black
```
christophlohrmann, line 17: Ci still uses black, right?

christophlohrmann (approved):

christophlohrmann:
thanks for the cleanup

SamTov (approved):
As long as test are passing, I am very happy with the cleanup

dev-requirements.txt
PythonFZ: CI uses black, ruff, isort and everyting else that is added to the .pre-commit-config.yaml

PythonFZ:
> As long as test are passing, I am very happy with the cleanup

I made no changes to the actual code.

